PR #7, opened 2015-11-09
- set BOOST_TEST_CATCH_SYSTEM_ERRORS=no

aschnell:


shundhammer:
LGTM

jreidinger:
just note that commit message should say why we do something, not that we do something. If in future we check why this line is here, we see only this commit message.

PR #13, opened 2015-11-13
- added to_device_of_type template function

aschnell:


storage/Devices/DeviceImpl.h
```diff
@@ -21,6 +21,32 @@ namespace storage
...
 
 
+    template <typename Type>
+    Type* to_device_of_type(Device* device)
+    {
+	static_assert(!is_const<Type>::value, "Type must not be const");
+
+	Type* tmp = dynamic_cast<Type*>(device);
+	if (!tmp)
+	    ST_THROW(DeviceHasWrongType("device has wrong type"));
```
mvidner: Wouldn't it be more informative to add the actual type?
```cc
ST_THROW(DeviceHasWrongType("device has wrong type, expected a " + typeid(Type).name()));
```

aschnell: Unfortunatelty that gives an error message like 'expected a N7storage10FilesystemE'.

So traits will be needed to get nice names.

Apart from that we can even go further:

```ST_THROW(DeviceHasWrongType(sformat("device has wrong type, expected '%s' but seen '%s'", typeid(Type).name(), device->get_impl().get_classname())));```

Comment (jreidinger) [inline storage/Devices/DeviceImpl.h]: To be honest, I expect such kind of exception constuct message at it own. Something like
```
ST_THROW(DeviceHasWrongType(Type, device))
```
and all that typeid, getting classname should be hidden inside DeviceHasWrongType constructor. just my 2c

jreidinger: and thinking about it probably I  will switch argument so it reflect name ( first device, second arg Type)

aschnell: Making the exception class a templated class visible in the public header files does not look like a good idea to me. That would likely cause trouble with some or all binding target languages.

But the two class names can be passed to the exception. Still, traits are needed and that is a bit more work.

jreidinger:
looks better. LGTM

PR #14, opened 2015-11-16
- added logger control functions to new api

aschnell:


storage/Logger.h
```diff
@@ -0,0 +1,33 @@
...
+namespace storage
+{
+
+    class Logger
+    {
+    public:
+
+	virtual ~Logger() {}
+
+	virtual bool test(int level);
```
jreidinger: what is purpose of this method? it always return true.

jreidinger:
in general I think it really helps if new API for logger have documentation stating what is its intention and intention of its methods.

storage/Logger.cc
```diff
@@ -0,0 +1,60 @@
...
+#include <iostream>
+
+#include "storage/Logger.h"
+#include "storage/Utils/AppUtil.h"
+
+
+namespace storage
+{
+
+    Logger* my_logger = nullptr;
```
shundhammer: Add 'static' ? This does not need to be exposed to the outside world.

storage/Logger.h
```diff
@@ -0,0 +1,44 @@
...
+{
+
+    /**
+     * The Logger class.
+     */
+    class Logger
+    {
+    public:
+
+	virtual ~Logger() {}
```
shundhammer: Even if it's just as empty as the dtor, a simple ctor would be nice.

storage/Utils/AppUtil.cc
```diff
@@ -368,13 +374,33 @@ bool isNfsDev( const string& dev )
...
+				  content.substr(pos1, pos2 - pos1));
 		if (pos2 == string::npos)
 		    break;
 		pos1 = pos2 + 1;
 	    }
 	}
+	else
+	{
+	    // legacy
+	    CallbackLogDo pfc = getLogDoCallback();
```
shundhammer: pfc? Private First Class? ;-)

Maybe use something more speaking like, say, app_logger_callback() ? it's only used twice in the function, and that name change would make quite some difference when trying to understand the code.

storage/Logger.cc
aschnell: OK.

storage/Utils/AppUtil.cc
aschnell: That is old code which will be removed when we can drop the compatibility layer.

shundhammer:
LGTM

PR #15, opened 2015-11-17
- added one more 'udevadm settle' call

aschnell:


storage/SystemInfo/CmdUdevadm.cc
```diff
@@ -37,6 +37,13 @@ namespace storage
     CmdUdevadmInfo::CmdUdevadmInfo(const string& file)
 	: file(file)
     {
+	// Without emptying the udev queue 'udevadm info' can display old data
+	// or even complain about unknown devices. Even during probing this
+	// can happen since e.g. 'parted' opens the disk device read-write
+	// even when all parted commands are read-only, thus triggering udev
+	// events. So always run 'udevadm settle'.
+	SystemCmd(UDEVADMBIN " settle --timeout=20");
```
shundhammer: I recommend a #define for that "settle --timeout=20" string.

aschnell: OK.

shundhammer:
LGTM

storage/SystemInfo/CmdParted.cc
```diff
@@ -39,7 +39,7 @@ namespace storage
     Parted::Parted(const string& device)
 	: device(device), label(PtType::PT_UNKNOWN), implicit(false), gpt_enlarge(false)
     {
-	SystemCmd cmd(PARTEDBIN " -s " + quote(device) + " unit cyl print unit s print", SystemCmd::DoThrow );
+	SystemCmd cmd(PARTEDBIN " -s " + quote(device) + " unit cyl print unit s print", SystemCmd::DoThrow);
```
kobliha: Please always use long form, so `--script` for `-s`. These are programs that need to be easy-to-read, no need to use shortcuts that are good only for typing faster. Thanks!

PR #17, opened 2015-11-19
- various minor improvements

aschnell:


storage/Devices/Device.cc
```diff
@@ -158,6 +188,26 @@ namespace storage
...
+
+
+    vector<Device*>
+    Device::get_ancestors(bool itself)
+    {
+	Devicegraph* devicegraph = get_impl().get_devicegraph();
+	Devicegraph::Impl::vertex_descriptor vertex = get_impl().get_vertex();
+
+	return devicegraph->get_impl().getDevices<Device>(devicegraph->get_impl().ancestors(vertex, itself));
+    }
```
jreidinger: hmm, it is not DRY and quite error prone to remember to modify both. I think we are not alone who face it and I am right. Here is hint how to do not without duplication - http://stackoverflow.com/questions/123758/how-do-i-remove-code-duplication-between-similar-const-and-non-const-member-func

aschnell: Did you try it? It's not so easy with vector<X*> vs. vector <const X*>.

mvidner:
LGTM

PR #18, opened 2015-11-19
- make Actiongraph part of public interface

aschnell:
Much code moved from Actiongraph.cc to ActiongraphImpl.cc. Otherwise many renamed from Actiongraph to Actiongraph::Impl. The rest is simple.


shundhammer:
LGTM

storage/Action.h
```diff
@@ -24,10 +24,10 @@ namespace storage
 
 	    virtual ~Base() {}
 
-	    virtual Text text(const Actiongraph& actiongraph, bool doing) const = 0;
-	    virtual void commit(const Actiongraph& actiongraph) const {} // = 0; TODO
+	    virtual Text text(const Actiongraph::Impl& actiongraph, bool doing) const = 0;
```
mvidner: I think this is wrong, as an Impl class has escaped to the public interface in this file.

mvidner: I have checked that this is indeed the only file where it happens:
```sh
grep ::Impl storage/*h storage/*/*h | grep -v Impl.h:
```

aschnell: Action.h is not part of the public interface. Maybe it will be in the future (see two TODOs in the code).

PR #25, opened 2015-11-27
- various improvements

aschnell:


integration-tests/delete-partition.py
```diff
@@ -22,5 +22,6 @@
 
 print staging
 
+storage.calculate_actiongraph()
```
mvidner: Wouldn't it be better if `commit` detected that the action graph was not calculated yet and did it automatically? Or am I understanding the API wrong?

aschnell: commit could calculate the actiongraph if it was never calculated before. But it does not know when to recalculate it after it was invalidated. That is the task of the user (see Storage.h).

bindings/storage.i
```diff
@@ -95,6 +95,7 @@
 %include "../../storage/Utils/HumanString.h"
 %include "../../storage/Utils/Region.h"
 %include "../../storage/Utils/Remote.h"
+%include "../../storage/Utils/Logger.h"
```
jreidinger: is there any ordering rule for includes? for alphabetical one it should be a bit upper

aschnell: Wrong ordering can cause problem for swig. Here I was just grouping subdirectories.

storage/Devicegraph.cc
```diff
@@ -43,13 +44,15 @@ namespace storage
 
 
     WrongNumberOfParents::WrongNumberOfParents(size_t seen, size_t expected)
-	: Exception(sformat("wrong number of parents, seen '%zu', expected '%zu'", seen, expected))
+	: Exception(sformat("wrong number of parents, seen '%zu', expected '%zu'", seen, expected),
+		    Silencer::is_any_active() ? DEBUG : WARNING)
```
jreidinger: I always see that silencer changed WARNING to DEBUG. Maybe it make sense to directly return level? like `Silencer::log_level` ?

aschnell: Indeed currently silencer always changes from WARNING to DEBUG but it might also change ERROR to DEBUG or ERROR to WARNING. The future will tell.

storage/Devices/FilesystemImpl.cc
```diff
@@ -396,14 +396,26 @@ namespace storage
     Text
     Filesystem::Impl::do_umount_text(const string& mountpoint, bool doing) const
     {
-	return sformat(_("Unmount %1$s"), get_displayname().c_str());
+	const BlkDevice* blkdevice = get_blkdevice();
+
+	return sformat(_("Unmount %1$s at %2$s"), blkdevice->get_name().c_str(),
+		       mountpoint.c_str());
```
jreidinger: I miss here what is purpose of `doing` variable?

aschnell: That parameter will be used to distinguish whether the text is used in the action overview ("Unmount ...") or during committing ("Unmounting ...") (just like in the old libstorage).

storage/Utils/Logger.cc
```diff
@@ -0,0 +1,98 @@
...
+
+
+    void
+    Silencer::turn_off()
+    {
+	if (!active)
+	    return;
+
+	active = false;
+	--count;
```
jreidinger: I do not get logic with counting. I expect turn_off do not set active false if count is bigger then one. Accoring to code it looks like count xor active flag is useless.
I do not get whole logic of it. As count cannot be bigger then one ( as it is guarded by active flag ). So what is purpose of that? Using only active flag ( when you want current behavior that on,on,off result in off ) or count ( when you want counting like on,on,off to be still on till next off ).

aschnell: count is shared between all Silencer objects and is non-zero if any Silencer is active.

storage/Devicegraph.cc
jreidinger: I just would like to know contract. E.g. if it always change to DEBUG then it can have method like Silence.silence(WARN) which return WARN or DEBUG depending if enabled.

storage/Utils/Logger.cc
jreidinger: ok, so there can be multiple silencer instance and if any instance is enabled it will be silenced, but one instance can be only on or off. Then it is fine, just probably not thread safe, but I am not sure what is libstorage contract regarding thread safety.

aschnell: There is no contract regarding thread safety.

storage/Devicegraph.cc
shundhammer: The purpose of that Silencer class is to prevent Storage::Exception subclasses from logging when I know that there might be exceptions, but I am watching out carefully for them, catch them and then act accordingly.

The general idea is to be able to react to special situations without flooding the log with warnings that there were exceptions. The calling code has to take care about it. This is the contract: I take responsibility for what's going to happen and you just shut up for a while - until I tell you otherwise.

silencer = ::Storage::Silencer.do
  begin
    partitions = disk.partition_table.partitions
  rescue ::Storage::Exception => ex
    log.info("No partition table on #{disk}")
  end
end

For that do..end block, Storage exceptions will be silent. They will not flood the log. That's all that class does. And that's the underlying contract.

jreidinger: so it probably make sense to always change to DEBUG. So call like `Silence.level(WARN)` that change to DEBUG make sense. It can also help in future if we decide we want to silence it to NONE, then it change in one place.

aschnell: Currently it makes some exception silent, e.g. DeviceHasWrongType but not NullPtrException.
And some exception log ERROR and some WARNING. We will have to see how this proves in the future, so currently no contract.

jreidinger: OK, it is not so critical. I just see it as example of "tell, not ask" principle. Now you need to ask silencer if it is active so ask...for tell it will be more give me log level. So responsibility to choose silenced log level is on silencer and not user.

jreidinger:
LGTM

storage/Devicegraph.cc
shundhammer: Do you seriously want to distribute such a setting all over the place? Because that's what's going to happen if we tell the Silencer in 4223423 locations what log level to use.

Bad idea.

jreidinger: now all over place is distributed that we silence from WARN to DEBUG. If we want to silence to given level, then it make sense for me to include target level for silence in silencer and not all over code. So change will be only that it say log level it want initially use and silencer decide if it want to silence it to silence level. I do not see how it can be worse then now.
instead of `Silencer::is_any_active() ? DEBUG : WARNING` will be `Silencer::silence(WARNING)` which return WARNING or DEBUG depending if silencer is on or off

PR #27, opened 2015-11-30
Prevent crash upon shutdown (object destruction sequence)

shundhammer:


aschnell:
Please add a TODO in the dtor since we have to tackle that problem.

jreidinger:
LGTM

aschnell:
LGTM

PR #30, opened 2015-12-04
- added overall documentation

aschnell:


doc/autodocs/doxygen.conf
```diff
@@ -17,7 +17,7 @@ INPUT = ../../storage ../../storage/Devices ../../storage/Holders		\
 FILE_PATTERNS = Actiongraph.h BlkDevice.h Device.h Devicegraph.h Disk.h		\
 	Encryption.h Environment.h Ext4.h Filesystem.h Gpt.h Holder.h		\
 	HumanString.h Msdos.h Partition.h PartitionTable.h Region.h Storage.h	\
-	StorageInterface.h Subdevice.h Swap.h User.h Logger.h LvmVg.h LvmLv.h
+	Subdevice.h Swap.h User.h Logger.h LvmVg.h LvmLv.h
```
mvidner: Yes, overall it is better to exclude the legacy API from the docs. It is only a pity that this also excludes the enums that will continue to be used in the new API: https://github.com/aschnell/libstorage-bgl-eval/commit/125cec13263a6248c04a5fb5fe080b862c02b883#diff-f0e07c09f56aca9d0cb9dda4d5e794cb . Maybe we should split this one into 2 files?

mvidner:
Just one suggestion. LGTM.

doc/autodocs/doxygen.conf
aschnell: I see. Depending on the future of the legacy interface we can split the enums into a separate file.

PR #32, opened 2015-12-07
Add ruby remote probe example

tgoettlicher:


bindings/ruby/examples/remote-probe.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  stdout.read.each_line { |line| ret.stdout << line.rstrip }
+  stderr.read.each_line { |line| ret.stderr << line.rstrip }
+
+  ret.exit_code = wait_thr.value.to_i
+
+  return ret
+end
+
+class MyRemoteCallbacks < Storage::RemoteCallbacks
```
aschnell: Maybe there should be a call to the constructor of the parent class Storage::RemoteCallbacks:

```
def initialize
    super
end
```

jreidinger: It is done automatic if no initialize is present. Just try it in irb:
```
irb(main):001:0> class A; def initialize(a); puts "Hey #{a}";end;end
=> :initialize
irb(main):002:0> class B < A;end
=> nil
irb(main):003:0> B.new("Arvin")
Hey Arvin
=> #<B:0x00000001b887f0>
```

aschnell: OK.

aschnell:
LGTM

PR #43, opened 2015-12-23
- define sorting of results

aschnell:


storage/DevicegraphImpl.h
```diff
@@ -111,58 +111,123 @@ namespace storage
...
+	    vector<Type*> ret = get_devices_of_type<Type>();
+
+	    sort(ret.begin(), ret.end(), compare);
+
+	    return ret;
+	}
+
 
-		ret.push_back(device);
+	template <typename Type, typename Pred>
```
shundhammer: Use a default parameter for pred (a TruePred that always returns true) and reduce the number of template functions to one of each kind?

aschnell: That would make all uses to the not _if functions more complicated (passing extra parameter). Since the predicate is not the last parameter a default value is not possible.

lslezak:
Small <sub><sup>LGTM</sup></sub> as I'm not a C++ expert and this is a quite advanced code for me...

PR #52, opened 2016-01-15
- added actions to handle /etc/mdadm.conf

aschnell:
These are just the Action classes to handle /etc/mdadm.conf. The concrete functions to handle /etc/mdadm.conf will follow later in another pull request.


shundhammer:
LGTM

PR #53, opened 2016-02-18
Expose get_geometry() / set_geometry() to API

shundhammer:
This is neccessary for setting up faked device graphs from YaML.


aschnell:
LGTM

PR #55, opened 2016-02-23
- added and use LogfileLogger

aschnell:
This is in preparation to drop legacy code, in particular old logging functions.

### DIFF
--- a/storage/Utils/Logger.cc
+++ b/storage/Utils/Logger.cc
@@ -35,8 +35,11 @@ namespace storage
 
     class StdoutLogger : public Logger
     {
+    public:
+
 	virtual void write(LogLevel log_level, const std::string& component, const std::string& file,
 			   int line, const std::string& function, const std::string& content) override;
+
     };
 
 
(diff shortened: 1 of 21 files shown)


storage/Utils/Logger.cc
```diff
@@ -58,6 +61,41 @@ namespace storage
...
+			   int line, const std::string& function, const std::string& content) override;
+
+    };
+
+
+    void
+    LogfileLogger::write(LogLevel log_level, const std::string& component, const std::string& file,
+			 int line, const std::string& function, const std::string& content)
+    {
+	FILE* f = fopen("/var/log/libstorage.log", "ae");
```
kobliha: So, this is some kind of a constant, isn't it?

aschnell: Yes, the name and access mode are constants. The name is on purpose a constant and the complete class is only for examples and test-cases. So I did not add functions to change the name.

shundhammer: Since this is only used once (in stark contrast to previously where it was in each of the examples), I wouldn't move it to a #define in this case; this would just scatter the information around (once for the #define, once for the fopen() call where that #define'd symbol would be used).

shundhammer:
LGTM

PR #56, opened 2016-02-24
- drop legacy interface

aschnell:


storage/Devices/Encryption.h
```diff
@@ -8,6 +8,11 @@
 namespace storage
 {
 
+    enum class EncryptType {
```
shundhammer: EncryptType sounds wrong. How about EncryptionType ?

shundhammer:
Other than that one issue, LGTM.

storage/Devices/Encryption.h
aschnell: Yes, indeed. EncryptType is still from the legacy interface.

PR #67, opened 2016-03-08
- reduce code duplication

aschnell:


testsuite/helpers/TsCmp.h
```diff
@@ -61,10 +60,20 @@ namespace storage
...
+	 * devicegraph and verifies the resulting actiongraph. Additionally it
+	 * generates a picture of the actiongraph.
+	 */
+	TsCmpActiongraph(const string& name);
+
+	TsCmpActiongraph(const Actiongraph& actiongraph, const Expected& expected);
 
     private:
 
+	void doit(const Actiongraph& actiongraph, const Expected& expected);
```
shundhammer: doit() is not a good name for anything... what does it do? Let's try to find a more speaking name.

testsuite/helpers/TsCmp.cc
```diff
@@ -51,19 +56,47 @@ namespace storage
...
+
+
+    TsCmpActiongraph::TsCmpActiongraph(const Actiongraph& actiongraph, const Expected& expected)
+    {
+	doit(actiongraph, expected);
+    }
+
+
+    void
+    TsCmpActiongraph::doit(const Actiongraph& actiongraph, const Expected& expected)
```
shundhammer: doit() is not a good name for anything... what does it do? Let's try to find a more speaking name.

shundhammer:
If we can find better names for those doit(), the rest LGTM.

PR #72, opened 2016-03-10
- work on unit tests

aschnell:


shundhammer:
Rubocop will complain about the empty lines, but aside from that, LGTM.

aschnell:
Rubocop does not mess around here.

PR #74, opened 2016-03-11
- detect reducing/extending of devices

aschnell:


storage/Devices/DeviceImpl.cc
```diff
@@ -140,6 +140,40 @@ namespace storage
     void
     Device::Impl::add_modify_actions(Actiongraph::Impl& actiongraph, const Device* lhs) const
     {
+	add_reallot_actions(actiongraph, lhs);
+    }
+
+
+    void
+    Device::Impl::add_reallot_actions(Actiongraph::Impl& actiongraph, const Device* lhs) const
```
shundhammer: Some comments would be very nice in this function. I have no idea what it does; it's not obvious from its name, and the purpose of those set operations does not become clear from reading the code.

storage/Devices/MdImpl.cc
```diff
@@ -410,6 +408,93 @@ namespace storage
...
+    }
+
+
+    void
+    Md::Impl::do_reallot(ReallotMode reallot_mode, const BlkDevice* blk_device) const
+    {
+	switch (reallot_mode)
+	{
+	    case ReallotMode::REDUCE:
+		do_reduce(blk_device);
```
shundhammer: Better:

    return do_reduce(blk_device);

also for the other case,
then get rid of that "default" branch and move the error handling for unknown enum values outside of the switch statement. Then the compiler can catch any unhandled enum value if there is ever a new one.

storage/Devices/MdImpl.cc
```diff
@@ -410,6 +408,93 @@ namespace storage
...
+	// print" readds the device to the md if the signature is still
+	// valid. Thus remove the signature.
+	blk_device->get_impl().wipe_device();
+    }
+
+
+    void
+    Md::Impl::do_extend(const BlkDevice* blk_device) const
+    {
+	string cmd_line = MDADMBIN " --add " + quote(get_name()) + " " + quote(blk_device->get_name());
```
shundhammer: move out that common block to a separate function? too much duplication here for my taste.

storage/SystemInfo/ProcMdstat.cc
```diff
@@ -128,18 +128,20 @@ namespace storage
 	    else
 		d = normalizeDevice(tmp);
 
+	    Device device(d);
```
shundhammer: Mark this class as ProcMdstat::Device for clarity to avoid confusion with Storage::Device?

integration-tests/md/extend.py
```diff
@@ -0,0 +1,30 @@
...
+from storageitu import *
+
+
+set_logger(get_logfile_logger())
+
+environment = Environment(False)
+
+storage = Storage(environment)
+
+staging = storage.get_staging()
```
mvidner: A staging what? A staging graph? A staging action?

integration-tests/md/extend.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/python
+
+# requirements: md raid /dev/md0 with partitions sdb[1-4] and unused partition sdb5
```
mvidner: Suppose I learn how to set up a proper environment for this test.
I still have no idea how to tell if the test passes or not. If it crashes, it failed. But beyond that?

integration-tests/md/extend.py
aschnell: The staging devicegraph since Storage::get_staging() returns a Devicegraph. Do you prefer ```staging_devicegraph = storage.get_staging_devicegraph()```?

integration-tests/md/extend.py
aschnell: Indeed, but as long as I don't have a framework where hundreds of such tests can easily be run automatically it's difficult to add code that checks the result. And I consider comments in human language as to inaccurate.

storage/SystemInfo/ProcMdstat.cc
aschnell: I can remove the variable completely.

storage/Devices/MdImpl.cc
aschnell: I could move the content of do_reduce and do_extend into do_allot but for my taste that does not make the workflow better understandable.

storage/Devices/DeviceImpl.cc
aschnell: Added some documentation.

storage/Devices/MdImpl.cc
aschnell: return the value of a function returning void looks strange to me although likely technically correct. So I will add ```return;``` instead.

shundhammer: Yes, that's also ok.

shundhammer:
LGTM now.

integration-tests/md/extend.py
mvidner: `get_staging_devicegraph()` would be more readable for me.  In C++ code it would be probably obvious, with explicitly declared types.
https://github.com/openSUSE/libstorage-ng/blob/dd5d8cfa5220b570e188f4e18621287ee15c5c10/storage/Storage.h#L113

PR #76, opened 2016-03-15
- added missing include (for whatever version travis/jenkins is using)

aschnell:


storage/Devices/BlkDeviceImpl.cc
```diff
@@ -1,5 +1,7 @@
 
 
+#include <iostream>
```
mvidner: https://ci.opensuse.org/job/libstorage-ng-master/7/console
But that failure is because of a ``` cout << cmd_line << endl;```. Do we really want to write to cout in the library?

aschnell: Yes, for the time being this is intended and used frequently.

shundhammer:
LGTM

PR #77, opened 2016-03-16
- added detailed flags for write_graphviz()

aschnell:


wfeldt:
lgtm

btw. would it have hurt to implement '||' as well?

aschnell:
I didn't see a use-case for ```||```.

PR #79, opened 2016-03-21
- removed now unneeded scope operator

aschnell:
The scope operator was needed when there were different Environment classes in the storage and storage_legacy namespace.


shundhammer:
LGTM

aschnell:
Thanks.

PR #81, opened 2016-03-22
- do not create devicegraph without storage object

aschnell:
Motivation is to solve the TODO at https://github.com/openSUSE/libstorage-ng/blob/master/storage/Devicegraph.h#L90.


mvidner:
LGTM

PR #82, opened 2016-03-29
swig bindings for Storage::Arch

wfeldt:
First try on getting bindings for Storage::Arch.

Not sure at all about the placement of that `#undef "_"`.

### DIFF
--- a/storage/SystemInfo/Arch.cc
+++ b/storage/SystemInfo/Arch.cc
@@ -71,19 +71,19 @@ namespace storage
 	if (is_ppc())
 	{
 	    AsciiFile cpuinfo("/proc/cpuinfo");
-	    vector<string>::const_iterator it = find_if(cpuinfo.lines(), string_starts_with("machine\t"));
+	    vector<std::string>::const_iterator it = find_if(cpuinfo.lines(), string_starts_with("machine\t"));
 	    if (it != cpuinfo.lines().end())
 	    {
 		y2mil("line:" << *it);
 
-		string tmp1 = extractNthWord(2, *it);
+		std::string tmp1 = extractNthWord(2, *it);
 		y2mil("tmp1:" << tmp1);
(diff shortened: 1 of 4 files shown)


storage/Utils/AppUtil.h
```diff
@@ -34,6 +34,7 @@
 #include <map>
 #include <chrono>
 
+#undef _
```
aschnell: Try to replace the #include "storage/Utils/XmlFile.h" by #include <libxml/tree.h> in Arch.h.

aschnell:
You also have to update bindings/storage.deps.

storage/SystemInfo/Arch.cc
```diff
@@ -71,19 +71,19 @@ namespace storage
 	if (is_ppc())
 	{
 	    AsciiFile cpuinfo("/proc/cpuinfo");
-	    vector<string>::const_iterator it = find_if(cpuinfo.lines(), string_starts_with("machine\t"));
+	    vector<std::string>::const_iterator it = find_if(cpuinfo.lines(), string_starts_with("machine\t"));
```
aschnell: In .cc file you can use using std::string.

wfeldt: the problem was that the generated storage_wrap.cxx does not know it

wfeldt: sorry, you're probably right; the problem rather was that the make dependencies required a complete rebuild

storage/Utils/AppUtil.h
wfeldt: no luck

aschnell:
Also please add Arch.h in doxygen.conf.

aschnell:
Arch.h should also be included in the devel package now, so add pkginclude_HEADERS in storage/SystemInfo/Makefile.am.

wfeldt:
replaced by https://github.com/openSUSE/libstorage-ng/pull/83

PR #85, opened 2016-04-11
- partition alignment, drop cylinders and geometry, change unit of sizes

aschnell:
You should not look at individual commits, there are intermediate mistakes. I recommend to first look at the changes in storage/Utils since those are the base for the rest.

Major API changes:
- *_size_k() -> *_size()
- Region::to_kb() -> to_bytes()
- Region::to_value() -> to_blocks()
- class Geometry dropped
- get/set_geometry() dropped
- region of class Partition is sector based
- class Topology added
- get/set_topology() added


storage/Utils/Topology.h
```diff
@@ -0,0 +1,94 @@
...
+	Topology& operator=(const Topology& topology);
+
+	long get_alignment_offset() const;
+	void set_alignment_offset(long alignment_offset);
+
+	unsigned long get_optimal_io_size() const;
+	void set_optimal_io_size(unsigned long optimal_io_size);
+
+	/**
+	 *
```
shundhammer: Light green font on light green background doesn't give much contrast... ;-)

Looking forward for the description here. But it was communicated that this is WIP, so this will come later.

storage/Utils/TopologyImpl.cc
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+
+    unsigned long
+    Topology::Impl::calculate_grain() const
+    {
+	unsigned long grain = max(optimal_io_size, 1UL);
+
+	while (grain < 1024 * 1024)
+	    grain *= 2;
```
shundhammer: I was tempted to suggest to use

    grain <<= 1;

but then the compiler very likely does this kind of simple optimization by itself, so better leave this as it is.

storage/Utils/TopologyImpl.cc
```diff
@@ -0,0 +1,155 @@
...
+	    }
+	}
+
+	ret += alignment_offset_in_blocks;
+
+	return ret;
+    }
+
+
+    // TODO aligned region can be non-existent, return length = 0 or exception?
```
shundhammer: IMHO exception. We are using exceptions for all kinds of much minor stuff lin libstorage-ng, and this is truly an exceptional situation.

shundhammer:
LGTM

doc/alignment.md
```diff
@@ -0,0 +1,61 @@
+
+Partition Alignment
+===================
+
+Unfortunately these days for optimal performance care must be taken that
+partitions are properly aligned, see e.g.
+https://en.wikipedia.org/wiki/Partition_alignment.
```
kobliha: It's very good to have a link here, statements like this have to be supported by more info, such as articles, research etc. Nice.

I just failed to find an exact part in that article (and those linked ones) that would explain that statement. Maybe I was not looking at the right place.

aschnell: At least in https://en.wikipedia.org/wiki/Advanced_Format:

> The translation process is more complicated when writing data that is either not a multiple of 4K or not aligned to a 4K boundary. In these instances, the hard drive must read the entire 4096-byte sector containing the targeted data into internal memory, integrate the new data into the previously existing data and then rewrite the entire 4096-byte sector onto the disk media. This operation, known as read-modify-write (RMW), can require additional revolution of the magnetic disks, resulting in a perceptible performance impact to the system user.

kobliha: Thanks. Then my personal test says that I'd expect mentioning at least a tiny bit of this paragraph and link to Advanced_Format

doc/style-guide.md
```diff
@@ -38,6 +38,7 @@ Filesystem.do_mount() and Filesystem.do_mount_text(). No other functions
 should have that naming schema.
 
 When variables or values with an underscores in the name are written to XML
-the underscores have to be replaced by dashed. So e.g. size_k is saved as
-<size-k>18</size-k>. In the logs the names should follow the same convention.
+the underscores have to be replaced by dashed. So e.g. chunk_size is saved as
```
shundhammer: "dashes", most likely, right?

storage/Devices/Disk.cc
```diff
@@ -35,6 +35,18 @@ namespace storage
 
 
     Disk*
+    Disk::create(Devicegraph* devicegraph, const string& name, const Region& region)
```
shundhammer: Do we have an alternative constructor for this that takes a size instead of a Region?

In this case, that Region will always start at 0. The sector size will normally be 512, so the only thing that is really variable is the size. Of course, disks with a different sector size will still need a Region to pass that sector size, but normally just the size will be sufficient.

storage/Devices/MdImpl.cc
```diff
@@ -291,25 +313,136 @@ namespace storage
...
+
+	if (real_chunk_size == 0)
+	    real_chunk_size = get_default_chunk_size();
+
+	// mdadm uses a chunk size of 64 KiB just in case the RAID1 is ever reshaped to RAID5.
+	if (md_level == RAID1)
+	    real_chunk_size = 64 * KiB;
+
+	int number = 0;
+	unsigned long long sum = 0;
```
shundhammer: Since this is used so many times here now, how about

    typedef unsigned long long DiskSize;

?
Not only would that be shorter to write, it would also clearly point out what kind of thing this is.

doc/style-guide.md
aschnell: Yes.

storage/Devices/MdImpl.cc
aschnell: Yes, but a simple typedef does not allow for nice features, e.g. function overloading (e.g. ostream<<).

Addind a class for that has been on the TODO list for quite some time: https://trello.com/c/ub28GvPD/56-dedicated-class-disksize

storage/Devices/Disk.cc
aschnell: I can add such a create function.

storage/Devices/MdImpl.cc
```diff
@@ -291,25 +313,136 @@ namespace storage
...
+
+	for (const BlkDevice* device : devices)
+	{
+	    // TODO handle spare
+
+	    unsigned long long size = device->get_size();
+
+	    // metadata for version 1.0 is 4 KiB block at end aligned to 4 KiB,
+	    // https://raid.wiki.kernel.org/index.php/RAID_superblock_formats
+	    size = (size & ~(0x1000 - 1)) - 0x2000;
```
shundhammer: Is this safe? 0x1000 is an integer constant, yet the calculations are based on the assumption that everything is unsigned long long. IMHO it would be a lot safer to user 0x1000ull here to ensure that type in the innermost statement.

storage/Devices/PartitionTableImpl.cc
```diff
@@ -199,32 +200,22 @@ namespace storage
...
+	bool is_extended_possible = is_primary_possible && extended_possible() && !has_extended();
+	bool is_logical_possible = has_extended() && num_logical() < (max_logical() - max_primary());
 
 	vector<PartitionSlot> slots;
 
 	vector<const Partition*> partitions = get_partitions();
 	sort(partitions.begin(), partitions.end(), compare_by_number);
 
-	if (all || !logical)
+	if (true)
```
shundhammer: Debugging artifact?

storage/Devices/MdImpl.cc
aschnell: Hmm, the bit representation and bit operations should be identical whether signed or unsigned is used. AFAIR I copied that line from libblkid.

storage/Devices/PartitionTableImpl.cc
aschnell: No, it's just to keep the code in a block like the code for logical partitions.

shundhammer:
LGTM now

PR #88, opened 2016-04-13
- use svg as output for dot in testsuite

aschnell:
png support is not available during build on some distributions, e.g. openSUSE 13.2.


ancorgs:
LGTM

PR #89, opened 2016-04-18
Added support to set label and uuid for swap devices

ancorgs:
set_uuid implemented at Filesystem level

do_set_uuid implemented only for swap. It would be easy to extend it for other filesystems using tune2fs.

### DIFF
--- a/storage/Devices/FilesystemImpl.cc
+++ b/storage/Devices/FilesystemImpl.cc
@@ -75,6 +75,13 @@ namespace storage
 
 
     void
+    Filesystem::Impl::set_uuid(const string& uuid)
+    {
+	Impl::uuid = uuid;
+    }
+
+
+    void
     Filesystem::Impl::set_mountpoints(const vector<string>& mountpoints)
     {
 	Impl::mountpoints = mountpoints;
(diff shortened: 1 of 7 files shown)


aschnell:
Good work, LGTM.

One thing is missing but I noticed it's also missing for the label. If the filesystem is not created but the uuid is changed, the action must also be inserted into the actiongraph. For an example have a look at Partition::Impl::add_modify_actions().

ancorgs:
Those are the caveats of copy&paste programming. You also replicate errors/omissions :smile:

I will add the scenario for both label and uuid

storage/Devices/FilesystemImpl.cc
```diff
@@ -378,6 +395,31 @@ namespace storage
 
 
     void
+    Filesystem::Impl::add_modify_actions(Actiongraph::Impl& actiongraph, const Device* lhs_base) const
+    {
+	const Impl& lhs = dynamic_cast<const Impl&>(lhs_base->get_impl());
+	vector<Action::Base*> actions;
+
+	if (get_type() != lhs.get_type())
```
aschnell: I don't think this can happen since a different filesystem object will have a different sid and thus the function will not be called. But the check doesn't hurt.

storage/Devices/FilesystemImpl.cc
```diff
@@ -378,6 +395,31 @@ namespace storage
...
+	{
+	    throw runtime_error("cannot change filesystem type");
```
aschnell: NP: In general the new ```ST_THROW(Exception(...))``` should be used.

aschnell:
LGTM.

PR #93, opened 2016-04-28
- allow to set minimal grain

aschnell:
Allow to set the minimal grain. Currently needed for testing the proposal for small MBR gaps.


wfeldt:
lgtm

PR #94, opened 2016-04-29
- allow to set minimal mbr gap

aschnell:
Allow to set the MBR gap. Currently needed for testing the proposal with too small MBR gaps.


wfeldt:
lgtm

PR #95, opened 2016-04-29
- initialise and log variable

aschnell:


wfeldt:
lgtm

mvidner:
LGTM.
The compiler warned about it, right?

aschnell:
No, not the compiler but coverity found the issue, see https://scan.coverity.com/projects/libstorage-ng.

aschnell:
The compiler can warn with the -Weffc++ option but that gives so many warnings that it's barely useable.

PR #96, opened 2016-05-20
Move the requirements doc to libstorage-ng

jsrain:
When included here, we should possibly remove it from the old libstorage repo

### DIFF
--- a/doc/yast-storage-requirements.md
+++ b/doc/yast-storage-requirements.md
@@ -7,7 +7,6 @@ This is about *libstorage* and the *yast-storage* Ruby code.
 * AI Lukas: ML: libstorage-ng@opensuse.org
 * AI PM: Clarify requirements on FAT partitions
 * AI lmb, Holgi: Claryfy requirements on filesystems (see below)
-* AI Arvin: clarify why: reate partitions on RAID (if there already is a partition table; cannot create a partition table if there is not one already
 * AI Architecture maintainers: Clarify which configurations of RAID are bootable, possibly under which conditions
 * AI: Veryfy support of resizing logical volumes for LVM
 * AI: Clarify why we do not support encryption of root filesystem, which tools would possibly need to be fixed
@@ -56,7 +55,7 @@ BIOS-managed raid is configured in BIOS (and Linux system detects it), Linux sof
 RAID levels:
 
    * **RAID 0**: striping; performance improvement due to parallel disk I/O; less fault tolerance than one disk
-   * **RAID 1**: mirroring; improved fault tolerance, same performance as one disk
+   * **RAID 1**: mirroring; improved fault tolerance, possibly better performance for reading
    * **RAID 5**: 3 or more disks with redundancy (for parity blocks): Good fault tolerance, improved performance (parallel I/O)
    * **RAID 6**: 4 or more disks; more redundancy than RAID 5, thus even better fault tolerance
    * **RAID 10**: striping + mirroring; combination of RAID 1 and RAID 0
@@ -64,7 +63,7 @@ RAID levels:
 Supported actions:
 
    * create RAID 0, 1, 5, 6, 10
-   * create partitions on RAID (if there already is a partition table; cannot create a partition table if there is not one already- AI: Arvin - why?)
+   * create partitions on RAID (if there already is a partition table; cannot create a partition table if there is not one already due to design limitation of old libstorage)
 
 Not supported by YaST storage - use dedicated tools instead:
 
@@ -209,7 +208,7 @@ Supported actions (for all filesystems, unless excluded in table above):
    * mount
    * unmount
 
-A file system can be located in a disk partition, in whole disk, on an MD array, on a logical volume
+Local file system can be located in a disk partition, in whole disk, on an MD array, on a logical volume
 
 
 ## Bootloader Constraints
@@ -336,7 +335,7 @@ Not supported by the automated proposal:
 
 ##Misc. Topics to be taken care of
 
-   * Support Systems with >> 800 hard disks
+   * Support Systems with >> 800 hard disks, provide still reasonable performance
    * btrfs subvolume handling
    * btrfs snapshots
    * btrfs rollback: for this we need to install into a btrfs snapshot already
--- a/doc/yast-storage-requirements.md
+++ b/doc/yast-storage-requirements.md
@@ -154,27 +154,27 @@ This covers storage devices, which are not accessible out-of-the-box and some op
 ### SystemZ Configuration
 (s/390 specific)
 
-       * DASDs (Direct Access Storage Devices) need to be activated for use with Linux with the *dasd\_configure* command
-       * DASDs are created/managed by the s/390 host system
-       * Ihno wants to provide a script which lists all disks
-       * zFCP
-       * AI: Provide specifics for zFCP devices
+   * DASDs (Direct Access Storage Devices) need to be activated for use with Linux with the *dasd\_configure* command
+   * DASDs are created/managed by the s/390 host system
+   * Ihno wants to provide a script which lists all disks
+   * zFCP
+   * AI: Provide specifics for zFCP devices
 
 ### iSCSI
 (Internet SCSI)
 
-       * networked storage using SCSI commands over TCP/IP
-       * using a separate set of commands and tools
-       * currently out of the scope of YaST storage, using a separate YaST module
-       * might need to be integrated into YaST storage some day
+   * networked storage using SCSI commands over TCP/IP
+   * using a separate set of commands and tools
+   * currently out of the scope of YaST storage, using a separate YaST module
+   * might need to be integrated into YaST storage some day
 
 ### FCoE
 (Fibre Channel Over Ethernet)
 
-       * networked storage using fibre channel over ethernet
-       * using a separate set of commands and tools
-       * currently out of the scope of YaST storage, using a separate YaST module
-       * might need to be integrated into YaST storage some day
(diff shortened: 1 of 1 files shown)


doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+   * Linux Software RAID / BIOS RAID (type dependent on BIOS):  
+       * mdadm RAID: https://en.wikipedia.org/wiki/Mdadm
+       * dmraid (Device Mapper RAID): https://en.wikipedia.org/wiki/Device_mapper ?
+
+BIOS-managed raid is configured in BIOS (and Linux system detects it), Linux softwre RAID must be configured in Linux
+
+RAID levels:
+
+   * **RAID 0**: striping; performance improvement due to parallel disk I/O; less fault tolerance than one disk
+   * **RAID 1**: mirroring; improved fault tolerance, same performance as one disk
```
aschnell: RAID 1 can have better performance for reading, see https://en.wikipedia.org/wiki/RAID.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+   * **RAID 5**: 3 or more disks with redundancy (for parity blocks): Good fault tolerance, improved performance (parallel I/O)
+   * **RAID 6**: 4 or more disks; more redundancy than RAID 5, thus even better fault tolerance
+   * **RAID 10**: striping + mirroring; combination of RAID 1 and RAID 0
+
+Supported actions:
+
+   * create RAID 0, 1, 5, 6, 10
+   * create partitions on RAID (if there already is a partition table; cannot create a partition table if there is not one already- AI: Arvin - why?)
```
aschnell: Concerning AI: Design limitation of old libstorage.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+Raid and boot partitions
+
+   * AI: Define which Raid configurations are bootable?
+
+### Volume Managers
+
+   * LVM
+   * Btrfs (volume manager features): not yet supported, needs FATE with usecases and technical information
+   * EVMS: legacy; no longer supported, drop
+   * AI: Find out if we need upgrade support of EVMS
```
aschnell: I don't think we need any upgrade support for EVMS since EVMS was already dropped for SLES11. See e.g. fate #305007 or https://bugzilla.suse.com/show_bug.cgi?id=848821#c18.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+
+Not supported by the automated proposal:
+
+   * RAID
+     * will work for HW RAID or BIOS RAID, which both can be seen as a single disk by proposal
+     * -> Options to be explored, so far we have no experience with proposal spread over multiple disks
+
+##Misc. Topics to be taken care of
+
+   * Support Systems with >> 800 hard disks
```
aschnell: Unspecific requirement.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+
+Supported actions (for all filesystems, unless excluded in table above):
+
+   * create
+   * resize (for some types, see above)
+   * /etc/fstab -handling (managing individual mount points, mount options)
+   * mount
+   * unmount
+
+A file system can be located in a disk partition, in whole disk, on an MD array, on a logical volume
```
aschnell: Not the case for network filesystems.

aschnell:
Some items belong to yast2-storage-ng, esp. the items about requirements for booting. Those are already documented there in a more detailed way.

aschnell:
The overall intension of the document is not clear to me:
- Are the requirements here also valid if there is not feature request?
- Is anything not mentioned here not supported?

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+   * LUKS multiple password slots handling
+   * change password
+   * set cipher (encryption method)
+
+### Multipath
+
+   * network storage devices (typically fibre channel) where multiple low-level kernel devices for the same storage device exist that have redundant connections to the machine 
+   * improved performance (parallel I/O over multiple channels) 
+   * fail safe against some I/O channels being blocked 
+   * special hardware that needs to be set up with special tools (outside the scope of YaST storage)
```
aschnell: Setup of multipath on the client is supported by libstorage and yast2-storage.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,346 @@
...
+| NTFS       |        |      | X      | AI for PM                                               |
+| tmpFS      |        |      |        | ignore in libstorag (except mounting)                   |
+| HFS        |        |      |        | drop                                                    |
+| HFS+       |        |      |        | drop                                                    |
+| NFS        |        | X    |        | installation on NFS as root is supported                |
+| ZFS        |        |      |        | Not supported                                           |
+| GFS2       | ?      |      |        | AI storage people                                       |
+| ceph       | ?      |      |        | AI storage people                                       |
+| ocfs2      | ?      |      |        | HA, AI lmb                                              |
+
```
aschnell: Swap should also be mentioned somewhere.

doc/yast-storage-requirements.md
jsrain: Do we have support for EVMS for upgrade in SLE12? If we don't, then we do not need it...

jsrain:
This document summarizes the requirements as discussed in NUE (you, Arvin, were not there) and as it went to the opensuse-storage ML. You are right that some parts do already exist in more details and others will follow, those should be replaced just with links. But I think that it still is worth having original version in the git history, as it is a bit between meeting minutes and docu.

doc/yast-storage-requirements.md
aschnell: Maybe there is old code but it's not a supported upgrade, see https://bugzilla.suse.com/show_bug.cgi?id=848821#c18.

aschnell:
Then the document should state that it's the summary of a meeting.

doc/yast-storage-requirements.md
jsrain: It is about cases when something special is necessary (hardware that needs special tool). If no special tools are necessary, it is in scope.

jsrain:
Added a comment about the origin of the document to the top.

doc/yast-storage-requirements.md
aschnell: OK, I always consider multipath hardware as special hardware.

doc/yast-storage-requirements.md
```diff
@@ -0,0 +1,348 @@
...
+* AI lmb, Holgi: Claryfy requirements on filesystems (see below)
+* AI Architecture maintainers: Clarify which configurations of RAID are bootable, possibly under which conditions
+* AI: Veryfy support of resizing logical volumes for LVM
+* AI: Clarify why we do not support encryption of root filesystem, which tools would possibly need to be fixed
+* AI: speak with different parties which requirements they have on automated proposal (Cloud, SES, SAP, ...)
+* AI: Define when dedicated boot partition is needed before merging into the 
+* AI: Find otu if EVMS support for upgrade is necessary (when was it last supported?)
+* AI: Provide specifics for zFCP devices
+* AI: Define requirements on /boot/zipl partition
+* AI: Provide details about specific tools for first sector writing, reasoning for two copies of kernel (why two?) for DASD
```
aschnell: This AI (the two copies thing) is not clear to me or maybe just wrong. What I know is that the first sector on some DASDs cannot be used generally, see https://bugzilla.suse.com/show_bug.cgi?id=894202#c30.

doc/yast-storage-requirements.md
jsrain: Well, some may be even more special :-)

doc/yast-storage-requirements.md
jsrain: Well, the AI is about finding out and documenting the limitations (actually, the bug is a good starting pointer). Latest when we start the DASD implementation, I would assume this AI to be clarified and documented for the purpose of implementing the right limitations.

shundhammer:
I am confused about this.

During that storage-ng kickoff meeting, we took that document we (Arvin and I) had written during last year's CSM workshop. It was a facts skeleton we used to move along during that meeting, kind of like an agenda.

Now there are small changes to that document (that are not visible in this PR because the original version was in another repository or had a different name), but there are still parts of the meeting minutes - such as the action items section. Does that really belong here? I don't think so.

Also, in the table of supported file systems, there are various "AI somebody" rows. Do we intend to maintain that status in this document? Do we have somebody who is in charge and will take care?

Maybe we should move those out to a separate table labeled "unclear" or "to be defined", even more so since it's about arcane file system types that are hardly relevant for what we are doing here (and there will always be new file system types, too).

jsrain:
The document was originally in the "old storage" repo - where it IMO does not belong to at all. It serves partially as meeting minutes, partially as the initial requirements documentation.

Long-term goal is form my POV to clarify all the missing stuff (AIs), which is not clear ATM, and have it documented in separate documents - and simultaneously remove the respective stuff from here (either completely, or replace them with links).

My main intention is to have the outcome of the discussion documented somewhere it is persistent (it is not the case of a mailing list) and do further refinements as stuff gets clarified and/or implemented. In the end (when we release), it is possible that this document gets empty (or removed).

kobliha:
It's just **moving from place A to B**. When we just had the old storage, it was basically the only place where to put it, then we created the new storage and as the document is about the new storage, it belongs there. Whether it contains / should contain some texts or not is rather a question for the next editor of the text.

PR #97, opened 2016-05-24
- removed perl bindings

aschnell:
The only potential user of the perl bindings was kiwi which was rewritten in Python. The perl-based kiwi is in maintanence mode.


shundhammer:
LGTM

PR #101, opened 2016-06-01
- work on LVM support

aschnell:
Completes minimal LVM probing support. Minimal means that only the LVM devices (LvmPv, LvmVg, LvmLv) with holders are added to the graph and only the volume group and logical volume names and the UUIDs are handled.

Saving and loading of objects works. A probing testcase is included.

For https://trello.com/c/vL3IRBRf/339-5-libstorage-ng-lvm-support-probing-part.


wfeldt:
lgtm

PR #103, opened 2016-06-09
- during probing find devices by major and minor number if find my name fails

aschnell:
This enables finding filesystems on lvm logical volumes.

libstorage-ng uses e.g. /dev/system/root as name for the logical volume while blkid reports e.g. /dev/mapper/system-root. Since a device can have so many names, e.g. /dev/system/root,
/dev/mapper/system-root, /dev/dm-0, /dev/block/254:0, /dev/disk/by-* the lockup via
major and minor number looks like the best approach.


wfeldt:
lgtm

PR #108, opened 2016-06-13
- use fallback lookup by major and minor number

aschnell:
Now filesystem the label and uuid on lvm logical volumes are detected properly.


shundhammer:
LGTM

PR #112, opened 2016-06-20
- probe size of lvm vg, lvm lv and luks

aschnell:


mvidner:
I barely understand the domain but it LGTM.

aschnell:
Thanks.

PR #116, opened 2016-06-27
- added bcache probing support

aschnell:
Adds [bcache](https://www.kernel.org/doc/Documentation/bcache.txt) probing support and some minor code improvements.

Adds the classes Bcache, derived from BlkDevice, and BcacheCset (bcache cache set), derived from Device, to the devicegraph. The picture shows the case of using one bcache cache set for two bcache devices.

display


mvidner:
The graph is pretty but even if I know what all the non-bcache things are it does not tell me what bcache is and what it is good for.

aschnell:
OK, [here is the link to the hackweek project](https://hackweek.suse.com/14/projects/1546) which includes links to bcache documentation.

storage/SystemInfo/DevAndSys.cc
```diff
@@ -122,6 +123,16 @@ namespace storage
     }
 
 
+    string
+    File::get_string() const
+    {
+	if (content.empty())
+	    ST_THROW(Exception("empty file"));
```
mvidner: what is wrong about an empty string?

aschnell: content is a ```vector<string>``` and if ```content``` is empty ```content.front()``` will fail. The string itself can be empty.

mvidner: duh, now I see it, thanks

storage/Devices/Bcache.h
```diff
@@ -0,0 +1,76 @@
...
+#define STORAGE_BCACHE_H
+
+
+#include "storage/Devices/BlkDevice.h"
+
+
+namespace storage
+{
+
+    //! A bcache device.
```
mvidner: A link to the docs here?

storage/Devices/Bcache.h
```diff
@@ -0,0 +1,76 @@
...
+    class Bcache : public BlkDevice
+    {
+    public:
+
+	static Bcache* create(Devicegraph* devicegraph, const std::string& dm_name);
+	static Bcache* load(Devicegraph* devicegraph, const xmlNode* node);
+
+	unsigned int get_number() const;
```
mvidner: Any number? OK, from the implementation I see
> "bcache3" -> 3

I would expect other BlkDevice to have a similar method and this being `override`

storage/Devices/BcacheCset.h
```diff
@@ -0,0 +1,80 @@
...
+
+
+#include "storage/Devices/Device.h"
+
+
+namespace storage
+{
+
+
+    //! A Bcache cache set.
```
mvidner: A short explanation of the relation between Bcache and BcacheSet? Like, which one is the SSD and which one hosts the FS on a spinning disk.

It is probably written in the upstream docs, but even the README manages to avoid the issue:
```
You'll need make-bcache from the bcache-tools repository. Both the cache device
and backing device must be formatted before use.
  make-bcache -B /dev/sdb
  make-bcache -C /dev/sdc
```

storage/Devices/BcacheCsetImpl.cc
```diff
@@ -0,0 +1,155 @@
...
+	Device::Impl::save(node);
+
+	setChildValue(node, "uuid", uuid);
+    }
+
+
+    bool
+    BcacheCset::Impl::is_valid_uuid(const string& uuid)
+    {
+	static boost::regex uuid_regex("[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}",
```
mvidner: is bcachecset the only class that handles uuids? if not then this should be a shared utility

storage/Devices/Bcache.h
aschnell: No, not all block devices have a number, e.g. a lvm logical volume or a luks device have non.
You would have to add a new class NumberedBlkDevice but that is likely not worth it.

storage/Devices/BcacheCsetImpl.cc
aschnell: I tought about that as well but at least LVM uses different UUIDs, e.g. "OMPzXF-m3am-1zIl-AVdQ-i5Wx-tmyN-cevmRn".

mvidner:
LGTM

aschnell:
Thanks.

storage/Devices/BcacheCsetImpl.cc
mvidner: (Regarding the difference I have found http://unix.stackexchange.com/questions/173722/what-is-the-uuid-format-used-by-lvm which is not complete but better than nothing.)

PR #120, opened 2016-07-05
- probe stripes and stripe-size for lvm lvs

aschnell:
Add support for probing the stripe count and stripe size of LVM logical volumes. Some minor moving of variables was required.


shundhammer:
LGTM

PR #122, opened 2016-07-11
- added convenience functions for lvm

aschnell:
For https://trello.com/c/PeiZJJrd/338-3-yast2-storage-ng-lvm-support-yaml-reader-writer.


storage/Devices/LvmVg.h
```diff
@@ -54,9 +63,43 @@ namespace storage
...
+	 * Removes a block device from the volume group. The physical volume
+	 * on the block device will the deleted.
+	 */
+	void remove_pv(BlkDevice* blk_device);
+
+	/**
 	 * Create a logical volume with name lv_name in the volume group.
 	 */
-	LvmLv* create_lvm_lv(const std::string& lv_name);
+	LvmLv* create_lvm_lv(const std::string& lv_name, unsigned long long size);
```
shundhammer: This is somewhat inconsistent: add_pv()/remove_pv() vs. create_lvm_lv()/delete_lvm_lv().

Since you need to create an instance of this LvmVg class anyway to do this, how about omitting that redundant _lvm in those calls?

testsuite/Devices/lvm-vg.cc
```diff
@@ -0,0 +1,153 @@
...
+    sda->set_size(16 * GiB);
+
+    LvmVg* lvm_vg = LvmVg::create(staging, "test");
+    lvm_vg->add_pv(sda);
+
+    BOOST_CHECK_EQUAL(staging->num_devices(), 3);
+    BOOST_CHECK_EQUAL(staging->num_holders(), 2);
+
+    BOOST_CHECK_EQUAL(sda->num_children(), 1);
+    BOOST_CHECK_EQUAL(lvm_vg->get_lvm_pvs().size(), 1);
```
shundhammer: Similar inconsistency with get_lvm_pvs()

storage/Devices/LvmVg.h
aschnell: Indeed, the naming is not consistent.

I'll change it to add_lvm_pv and remove_lvm_pv to be consistent. The function name should reflect the class name.

shundhammer:
LGTM now.

PR #130, opened 2016-08-08
Make libstorage-ng installable alongside libstorage

ancorgs:
Needed to get travis green https://travis-ci.org/yast/yast-installation/builds/150581997#L460


storage/Utils/Lock.cc
```diff
@@ -33,7 +33,7 @@
 #include "storage/Utils/Lock.h"
 
 
-#define LOCKDIR "/run/libstorage"
+#define LOCKDIR "/run/libstorage-ng"
```
aschnell: I don't think that having different lock files for the libraries is a good idea.

ancorgs: Then, how can we share the same lock file in both packages?

aschnell: AFAIR: Directories can be owned by several RPMs. The lock-file itself is generated if missing so it could be marked as ghost in both RPMs. On the other hand I didn't use RPM much in the past and definitely don't know all the packaging rules.

ancorgs:
@aschnell I double checked and the lock file was not conflicting. I removed the corresponding commit and checked that both packages can co-exist. Please re-review.

aschnell:
LGTM

Why do we need it?

ancorgs:
In order to make travis work for the storage-ng branch of yast2-installation we need to install both yast-storage-ng and many other packages in the test system. Some of those other packages end up depending on the old libstorage. So we need both libstorage living together.

PR #131, opened 2016-08-12
- fixed exceptions

aschnell:


ancorgs:
LGTM, but I would prefer not to merge it yet. Making libstorage-ng crash in my real hardware will be useful to verify the next upcoming fix (correct passing of exceptions), and this fix will fix that useful crash. :wink:

ancorgs:
LGTM

PR #132, opened 2016-08-12
Catch base exception in the bindings

ancorgs:
Tested manually in the installer. It fixes the observed crash.


aschnell:
Included in #131.

PR #136, opened 2016-08-17
Obsolete again old libstorage

ancorgs:
We moved .deb packages depending on libstorage-ng to their own OBS project and removed all explicit dependencies from the old stack. Thus, we don't need both generations to be installed at the same time anymore.


aschnell:
LGTM

PR #137, opened 2016-08-18
- add dependency for reallot actions

aschnell:
For https://trello.com/c/5beXKOT2/412-5-yast2-storage-ng-lvm-support-commit.

The main commit is https://github.com/openSUSE/libstorage-ng/commit/6b385a7230dc753bfaaeb150d38a2ab840b6777c and rather short. The rest deals with some renaming and unit tests.


ancorgs:
In the name of the tests only "reduce" and "extend" are used. But nothing indicates that it's about reducing and extending a VG. It will be a problem when we introduce tests for extending and reducing LVs. And also it makes harder to figure out what the test is about at a first glance.

aschnell:
The naming is no problem since libstorage-ng has a consistent terminology here: https://github.com/openSUSE/libstorage-ng/blob/master/storage/Devices/DeviceImpl.h#L57

ancorgs:
Oh, right. LGTM

PR #139, opened 2016-08-19
- moved calculating dependencies for creating partitions to PartitionTable class

aschnell:
Moved calculating dependencies for creating partitions to PartitionTable class. New code is simpler, more efficient and generates less strict dependencies. The sorting by partition number is also correct now.

Preparation for LVM commit action dependencies, see https://trello.com/c/5beXKOT2/412-5-yast2-storage-ng-lvm-support-commit.


ancorgs:
LGTM

PR #140, opened 2016-08-29
Implement dependencies between LVM actions

ancorgs:


storage/Action.cc
```diff
@@ -95,11 +99,17 @@ namespace storage
 			sid_t child_sid = actiongraph.get_devicegraph(LHS)->get_impl()[*vi2]->get_sid();
 
 			vector<Actiongraph::Impl::vertex_descriptor> tmp = actiongraph.actions_with_sid(child_sid, ONLY_LAST);
-			if (!tmp.empty())
-			    actiongraph.add_edge(tmp.front(), v);
+			if (!tmp.empty()) {
```
shundhammer: Please follow the same coding style as in the rest of that library: Opening curly braces go to the next line.

storage/Devices/LvmVgImpl.cc
```diff
@@ -520,4 +520,101 @@ namespace storage
...
+	vector<Actiongraph::Impl::vertex_descriptor> increase_lv_actions;
+
+	// Look for the relevant actions
+	for (Actiongraph::Impl::vertex_descriptor vertex : actiongraph.vertices())
+	{
+	    const Action::Base* action = actiongraph[vertex];
+
+	    if (action_is_my_reallot(action, actiongraph))
+		reallot_actions.push_back(vertex);
+	    else if (action_is_on_my_lv(action, actiongraph)) {
```
shundhammer: Coding style again: Please move the { to the next line.

storage/Action.cc
ancorgs: Oops. Rubocop made me lazy in checking this stuff. :-) I will fix it in the whole pull request.

storage/Devices/LvmVgImpl.cc
```diff
@@ -520,4 +520,101 @@ namespace storage
...
+		if (action_frees_vg_space(action)) {
```
shundhammer: ...and one more { to go to the next line...

storage/Devices/LvmVgImpl.cc
```diff
@@ -520,4 +520,101 @@ namespace storage
...
+		    decrease_lv_actions.push_back(vertex);
+		} else {
```
shundhammer: ...and again:

    }
    else
    {

shundhammer:
Besides the coding style issues, LGTM.

shundhammer:
LGTM now.

storage/Devices/LvmVgImpl.cc
```diff
@@ -520,4 +520,102 @@ namespace storage
...
     }
 
+
+    void
+    LvmVg::Impl::add_dependencies(Actiongraph::Impl& actiongraph) const
+    {
+	// First, all the operations removing or shrinking LVs
+	vector<Actiongraph::Impl::vertex_descriptor> decrease_lv_actions;
+	// Then a sync point
+	Action::Base* decrease_lv_sync = new Action::Create(get_sid(), true);
```
aschnell: Does Action::Base also work? These sync objects are not really of kind Action::Create.

storage/Devices/LvmVgImpl.cc
```diff
@@ -520,4 +520,102 @@ namespace storage
...
+
+	Actiongraph::Impl::vertex_descriptor sync2 = actiongraph.add_vertex(reallot_sync);
+	for (const Actiongraph::Impl::vertex_descriptor& vertex : reallot_actions)
+	{
+	    actiongraph.add_edge(sync1, vertex);
+	    actiongraph.add_edge(vertex, sync2);
+	}
+
+	for (const Actiongraph::Impl::vertex_descriptor& vertex : increase_lv_actions)
+	    actiongraph.add_edge(sync2, vertex);
```
aschnell: Does this work if reallot_actions is empty? Is there then a dependency from sync1 to sync2.

aschnell: Can the add_edge functionality be moved to a own function, like Actiongraph::Impl::add_chain(). Could take a vector of vectors of Actiongraph::Impl::vertex_descriptor.

Maybe the code can be reused for e.g. btrfs.

storage/Devices/LvmVgImpl.cc
ancorgs: No. I tried but Action::Base is an abstract class. Maybe we should make possible to create Base objects or (my preferred solution) have a Action::Sync instead of a boolean attribute.

storage/Devices/LvmVgImpl.cc
ancorgs: I don't think it will work. I will create a test case to make sure we don't skip that case.

ancorgs:
Added a test that proves @aschnell was right and the current code fails if no PV operations are done between shrinking and creating LVs. Pull request should turn red now.

storage/Devices/LvmVgImpl.cc
ancorgs: For the empty `reallot_actions`: test added and issue fixed in recent commits.

Adding edges moved to a reusable function, as requested.

storage/Devices/LvmVgImpl.cc
ancorgs: My code does not longer use sync points. But still I think we should improve this (in a different pull request).

@aschnell What about removing the `only_sync` from `Action::Base` and creating a `Action::Noop` class for syncing purposes? I found it less hacky and easier to debug problems (since `Action::Noop` would have a meaningful implementation of `text()`).

ancorgs:
Pull request updated to address @aschnell's concerns and suggestions. Also added dependencies for moving PVs between different VGs. Tests included for all scenarios.

storage/ActiongraphImpl.cc
```diff
@@ -130,6 +130,23 @@ namespace storage
...
 
 
+    void
+    Actiongraph::Impl::add_chain(const vector<vector<vertex_descriptor>>& vert_vectors)
+    {
+	vector<vector<Actiongraph::Impl::vertex_descriptor>> non_empty_vectors;
+	for (const vector<vertex_descriptor>& vector : vert_vectors)
+	    if (!vector.empty()) non_empty_vectors.push_back(vector);
+
+	if (non_empty_vectors.size() > 1)
```
shundhammer: NP:

    if (!non_empty_vectors.empty())

storage/ActiongraphImpl.cc
```diff
@@ -130,6 +130,23 @@ namespace storage
...
+	    for (size_t i = 1; i < non_empty_vectors.size(); ++i)
```
shundhammer: Hm... this has O(n^3) AFAICS. Probably n will always be pretty small, but can we be sure about that? What about S/390 that might have some hundred DASDs with up to 3 partitions on each? Will that hurt us?

storage/ActiongraphImpl.h
```diff
@@ -95,10 +95,26 @@ namespace storage
 
 	edge_descriptor add_edge(vertex_descriptor a, vertex_descriptor b);
 
+	/**
+	 * Adds several actions (with their corresponding vertex_descriptors) to
+	 * the graph, linking each other with edges (dependencies) in the given
+	 * order.
+	 */
```
shundhammer: +1 for function documentation!

storage/ActiongraphImpl.cc
ancorgs: No. We are checking if `> 1`, not ` > 0`.

storage/ActiongraphImpl.cc
ancorgs: Right now it's used for dependencies of operations within a LVM VG. The cost should be kind of obvious from the description, but I will add a note.

shundhammer: OK then; we shouldn't get so many of those that this might become a problem.

storage/ActiongraphImpl.cc
shundhammer: Right, my mistake.

shundhammer:
LGTM

storage/Devices/LvmVgImpl.cc
aschnell: In the beginning there was an Action::Nop but it also had problems. AFAIR even for Nop you need the information whether it's create or delete so that you can e.g. delete children before the parents - even if the delete action itself is a nop.

PR #142, opened 2016-09-09
- support to create, activate, deactivate and delete Luks

aschnell:
For https://trello.com/c/KDsLDxVQ/414-5-yast2-storage-ng-lvm-plain-devices-luks.

Individual commits are consistent.


integration-tests/luks/create-with-ext4.py
```diff
@@ -0,0 +1,30 @@
...
+
+environment = Environment(False)
+
+storage = Storage(environment)
+
+staging = storage.get_staging()
+
+sdb1 = Partition.find_by_name(staging, "/dev/sdb1")
+
+luks = sdb1.create_encryption("cr_test")
```
mvidner: "Create encryption" does not make much sense. Can we rename this to "Create encrypted device/volume/partition/whatever"?

aschnell: Why do is not make sense? The object being created is called Encryption so it is consistent with other functions, e.g. create_filesystem() or create_lvm_lv().

mvidner: > The object being created is called Encryption

That does not make sense either. Encryption is a concept or an activity, not an object. When I look at the [class declaration](https://github.com/aschnell/libstorage-ng/blob/33769c90ef09168229d61a8e2f55c40217b2e7ce/storage/Devices/Encryption.h#L43) I think we should replace `Encryption` by `EncryptionLayer`.

aschnell: To me that is just nit-picking and would make class and function names longer without a real benefit.

storage/Devices/DeviceImpl.cc
```diff
@@ -283,7 +290,35 @@ namespace storage
     void
     Device::Impl::do_delete() const
     {
-	throw std::logic_error("stub function called");
+	ST_THROW(LogicException("stub function called"));
```
shundhammer: How about declaring those methods as pure virtual so the compiler can generate code to catch this kind of thing?

integration-tests/luks/create-with-ext4.py
shundhammer: While it is true that "encryption" is really a concept, IMHO it is clearly not an activity (that would be the act of encrypting). But since you can perform actions on that thing, it is as much an object as a file system is an object: You can create it, destroy it, set its password. And with the password it also has data which is the other thing that constitutes an object in the OOP sense.

I'd really hate to mess with ugly and artificial sounding things like EncryptionLayer in this context. IMHO this would pretty much destroy the abstraction level we already have with block devices, disks, partition tables, partitions, file systems. I strongly vote for keeping "Encryption" as another one like those.

shundhammer:
Apart from the suggested virtual methods (where there might be a good reason against), LGTM.

storage/Devices/DeviceImpl.cc
aschnell: Some devices do not support do_delete, e.g. disks. So making do_delete pure virtual would just require to reimplement it in the derived classes and likely throw an exception there.

PR #144, opened 2016-09-13
- moved some classes to own files

aschnell:
Not related to a specific PBI but "Boy Scout Rule" development.

### DIFF
--- a/storage/Action.h
+++ b/storage/Action.h
@@ -25,7 +25,7 @@
 #define STORAGE_ACTION_H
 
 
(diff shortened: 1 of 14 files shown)


ancorgs:
+1 for the code reorganization.

But, since you are wearing the Boy Scout uniform already, it would be nice to document a little bit the new class: the meaning of `native` and `text`, the purpose of `UntranslatedText`... Just some words here and there.

ancorgs:
Thanks. LGTM.

PR #150, opened 2016-09-22
- temporarily mount filesystem for detect_resize_info and detect_content_info if required

aschnell:
Temporarily mount filesystem for detect_resize_info and detect_content_info if required. Uses RAII for excepetion safety and to avoid programming mistakes.

For https://trello.com/c/YTeaoJYo/411-3-storageng-do-real-check-for-the-windows-partition.

### DIFF
--- a/storage/Filesystems/FilesystemImpl.cc
+++ b/storage/Filesystems/FilesystemImpl.cc
@@ -226,16 +226,9 @@ namespace storage
 
 	// TODO only in real probe mode allowed
 
-	const BlkDevice* blk_device = get_blk_device();
+	EnsureMounted ensure_mounted(this, TmpMount::READ_ONLY);
 
-	blk_device->get_impl().wait_for_device();
-
-	// TODO filesystem must be mounted
-
-	if (mountpoints.empty())
-	    ST_THROW(Exception("filesystem must be mounted"));
-
-	StatVfs stat_vfs = detect_stat_vfs(mountpoints.front());
+	StatVfs stat_vfs = detect_stat_vfs(ensure_mounted.get_any_mountpoint());
 
 	ResizeInfo resize_info(true);
 	resize_info.min_size = stat_vfs.size - stat_vfs.free;
@@ -273,19 +266,12 @@ namespace storage
 
 	// TODO only in real probe mode allowed
 
-	const BlkDevice* blk_device = get_blk_device();
-
-	blk_device->get_impl().wait_for_device();
-
-	// TODO filesystem must be mounted
-
-	if (mountpoints.empty())
-	    throw;
+	EnsureMounted ensure_mounted(this, TmpMount::READ_ONLY);
 
 	ContentInfo content_info;
 	content_info.is_windows = false;
 	content_info.is_efi = false;
-	content_info.num_homes = detect_num_homes(mountpoints.front());
+	content_info.num_homes = detect_num_homes(ensure_mounted.get_any_mountpoint());
 
 	return content_info;
     }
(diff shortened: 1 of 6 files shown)


storage/Utils/FileUtils.h
```diff
@@ -54,14 +61,38 @@ namespace storage
...
     };
 
-}
 
+    class TmpMount : public TmpDir
+    {
+
+    public:
+
+	enum Mode { READ_ONLY, READ_WRITE };
```
mvidner: We could use `enum class` here, right?

aschnell: Yes, I can change that.

Since the enum is already visible only in the class namespace I did not consider it nessesary.

storage/Filesystems/FilesystemImpl.h
```diff
@@ -252,6 +253,32 @@ namespace storage
...
+
+    class EnsureMounted : boost::noncopyable
+    {
+
+    public:
+
+	/**
+	 * Ensures that the filesystem is mounted somewhere.
+	 *
+	 * The mode is not enforced.
```
mvidner: What does this mean? A "TODO, not implemented"? But we do pass `--read-only` to `mount`.

storage/Utils/FileUtils.h
```diff
@@ -54,14 +61,38 @@ namespace storage
...
 
+    class TmpMount : public TmpDir
+    {
+
+    public:
+
+	enum class Mode { READ_ONLY, READ_WRITE };
+
+	/**
+	 * Mounts device at TmpDir(path, name_template).
```
mvidner: We should add "throws on failure" and (also document that TmpDir does not throw if its destructor fails to remove the dir)

storage/Filesystems/FilesystemImpl.cc
```diff
@@ -940,4 +929,30 @@ namespace storage
...
+    }
+
+
+    string
+    EnsureMounted::get_any_mountpoint() const
+    {
+	if (tmp_mount)
+	    return tmp_mount->get_fullname();
+	else
+	    return filesystem->get_mountpoints().front();
```
mvidner: Could it happen that `filesystem->get_mountpoints()` changes between our constructor and this call?

storage/Filesystems/FilesystemImpl.h
aschnell: If the device is already mounted the mode is ignored since the class does not do a second mount.

storage/Filesystems/FilesystemImpl.cc
aschnell: That must not happen. Just like nobody must unmount the temporary mount.

storage/Utils/FileUtils.h
aschnell: Throw on failure is standard in the code. A throw in a destructor will likely terminate the program and is basically never used.

storage/Filesystems/FilesystemImpl.h
aschnell: And even if we do a second mount you cannot mount a filesystem once read-only and once read-write.

mvidner: Then wouldn't it be better to drop the Mode arguments and document that we cannot guarantee the mode anyway? That's certainly better for the current use case.

storage/Utils/FileUtils.h
mvidner: OK, that makes sense. I must refresh my C++.

PR #152, opened 2016-09-27
- removed some examples that were initial test programs

aschnell:
The example programs are also outdated, e.g. not creating a PartitionTable object or not using the latest API functions.


shundhammer:
LGTM

PR #162, opened 2016-10-19
- make logical partitions children of extended partition

aschnell:
For https://trello.com/c/D61iRjOo/424-5-yast2-storage-ng-make-logical-partitions-children-of-the-extended-partition plus the extra mile.

Individual commits are consistent.

yast2-storage-ng needs adaptation.


storage/Devices/MsdosImpl.cc
```diff
@@ -167,6 +168,84 @@ namespace storage
...
+		vector<const Partition*> logicals = get_logical_partitions(partition);
+		partitions.insert(partitions.end(), logicals.begin(), logicals.end());
+		break;
+	    }
+	}
+
+	return partitions;
+    }
+
+
```
ancorgs: Just asking from my C++ ignorance. There is no easy way to avoid the code duplication between `vector<Partition*> Msdos::Impl::get_partitions()` and `vector<const Partition*> Msdos::Impl::get_partitions() const`?

Same goes for `get_logical_partitions`

wfeldt:
lgtm

storage/Devices/MsdosImpl.cc
wfeldt: I don't think so.

ancorgs:
LGTM as well

PR #163, opened 2016-10-24
- added name for partition id

aschnell:
For https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi.


aschnell:
Yes, I considere the strings to be technical terms that do not need translation.

Translating them is also not easy, that would require the sformat function to accept Text parameters.

PR #166, opened 2016-10-26
Added LvmLv.get_all()

ancorgs:
Without this, `::Storage::LvmLv.all(devicegraph)` returns a list of all the devices. Quite unexpected.


aschnell:
LGTM

PR #169, opened 2016-10-28
- full support for boot, legacy_boot and pmbr_boot flags

aschnell:
Bonus for https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi.

### DIFF
--- a/integration-tests/partitions/create-partition.py
+++ b/integration-tests/partitions/create-partition.py
@@ -52,6 +52,8 @@ if not good_partition_slot:
 partition = partition_table.create_partition(good_partition_slot.name, good_partition_slot.region,
                                              type)
 
+#partition.set_boot(True)
+
 print staging
 
 commit(storage)
--- a/integration-tests/partitions/create-partition.py
+++ b/integration-tests/partitions/create-partition.py
@@ -52,8 +52,6 @@ if not good_partition_slot:
 partition = partition_table.create_partition(good_partition_slot.name, good_partition_slot.region,
                                              type)
 
-#partition.set_boot(True)
-
 print staging
 
 commit(storage)
--- a/storage/Devices/Partition.h
+++ b/storage/Devices/Partition.h
@@ -77,9 +77,18 @@ namespace storage
 	bool is_boot() const;
 
 	/**
-	 * Set the boot flag of the partition. Only supported on Msdos. Note:
-	 * The boot flag known in parted for Gpt sets the partition id to ESP
-	 * and is not a discrete flag.
+	 * Set the boot flag of the partition. Only supported on Msdos.
+	 *
+	 * Notes:
+	 *
+	 * 1. To be
+	 * [standard-conformant](https://en.wikipedia.org/wiki/Master_boot_record),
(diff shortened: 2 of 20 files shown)


integration-tests/partitions/create-partition.py
```diff
@@ -52,6 +52,8 @@
 partition = partition_table.create_partition(good_partition_slot.name, good_partition_slot.region,
                                              type)
 
+#partition.set_boot(True)
+
```
ancorgs: Why is this commented code left?

aschnell: Test code. I'll remove it.

storage/Devices/Partition.h
```diff
@@ -71,9 +71,28 @@ namespace storage
...
-	bool get_boot() const;
+	/**
+	 * Query the boot flag of the partition.
+	 */
+	bool is_boot() const;
+
+	/**
+	 * Set the boot flag of the partition. Only supported on Msdos. Note:
+	 * The boot flag known in parted for Gpt sets the partition id to ESP
+	 * and is not a discrete flag.
```
ancorgs: What do you mean with "is not a discrete flag"?... In general, I'm not sure if I understand the whole note.

aschnell: Partitions on GPT have no boot flag, "set <nr> boot on" with parted on GPT partitions only sets the partition type to EFI System Partition. That can also be done with "set <nr> esp on".

In general the id/flag handling in parted is very weird and I prefer to have a clear interface in the library.

storage/Devices/PartitionImpl.cc
```diff
@@ -269,6 +307,34 @@ namespace storage
...
+    Partition::Impl::set_boot(bool boot)
+    {
+	if (!is_msdos(get_partition_table()))
+	    ST_THROW(Exception("set_boot only supported for Msdos"));
+
+	if (boot && !Impl::boot)
+	{
+	    PartitionTable* partition_table = get_partition_table();
+
+	    for (Partition* partition : partition_table->get_partitions())
```
ancorgs: But it can be done with cfdisk. So it's possible. For example, in [wikipedia's article](https://en.wikipedia.org/wiki/Boot_flag) it says "some modern BIOS test if the bootflag of **at least one** partition is set". Which also seems to imply that is possible.

The fact that we use parted under the hood and that parted cannot do it is just an implementation detail. As long as there are common tools out there allowing it, we should document that our library doesn't.

storage/Devices/Partition.h
ancorgs: So please replace the current note with the first paragraph of your comment, which is way more clear (to me, at least).

storage/Devices/PartitionImpl.cc
aschnell: Well, then those programs are not standard-conformant: "Standard-conformant MBRs will allow only one partition marked active and use this as part of a sanity-check to determine the existence of a valid partition table." (https://en.wikipedia.org/wiki/Master_boot_record)

But I can add a sentence about this in the documentation.

ancorgs:
LGTM

PR #172, opened 2016-11-18
- added interface to query used features

aschnell:
The interface can be used to implement SaveUsedFs() and AddPackageList() in Storage.rb for storage-ng. See https://github.com/yast/yast-storage/blob/master/src/lib/storage/used_storage_features.rb for the implementation working on the target-map.

Also see https://trello.com/c/ZSnI5lF0/446-1-summarize-what-the-software-proposal-really-needs-from-storage.


ancorgs:
In general it looks good. But I miss some unit test. It's a completely new feature and it should be tested.

ancorgs:
LGTM now

PR #174, opened 2016-12-02
handle partition id on GPT

aschnell:
For https://trello.com/c/Mqq4FiIa/485-5-storageng-handle-partition-ids-for-gpt.

yast2-storage-ng needs adaption to the changed enums. Pull request will follow soon.


storage/Devices/PartitionImpl.cc
```diff
@@ -499,16 +580,71 @@ namespace storage
 
 	// TODO
```
ancorgs: Does this TODO still stands?

storage/Devices/PartitionImpl.cc
```diff
@@ -499,16 +580,71 @@ namespace storage
...
+		case ID_MICROSOFT_RESERVED:
+		    cmd_line += "msftres";
+		    break;
+	    }
+
+	    cout << cmd_line << endl;
+
+	    SystemCmd cmd(cmd_line);
+	    if (cmd.retcode() != 0)
+		ST_THROW(Exception("set partition id failed"));
```
ancorgs: NP: isn't it redundant to start and finish both branches of the script with basically the same lines of code (preparing cmd_line and executing it respectively)?

storage/Devices/PartitionTableImpl.h
```diff
@@ -75,6 +75,37 @@ namespace storage
...
+	/**
+	 * Returns whether the id is supported for the partition table. May
+	 * return true for ID_UNKNOWN but committing ID_UNKNOWN to disk is not
+	 * supported.
+	 */
+	virtual bool is_partition_id_supported(unsigned int id) const = 0;
+
+	/**
+	 * Returns whether the value of the partition id enum is useful and
+	 * should be include in e.g. action messages.
```
ancorgs: included (missing "d")

storage/Devices/PartitionImpl.cc
```diff
@@ -481,13 +518,57 @@ namespace storage
     Text
     Partition::Impl::do_set_id_text(Tense tense) const
     {
+	const PartitionTable* partition_table = get_partition_table();
+
 	string tmp = id_to_string(get_id());
 
-	if (tmp.empty())
-	    return sformat(_("Set id of partition %1$s to 0x%2$02X"), get_name().c_str(), get_id());
+	if (partition_table->get_impl().is_partition_id_value_meaningful())
```
ancorgs: Not sure if I got why we make a difference between meaningful and not meaningful ids here. With meaningful but empty ids (surprising scenario, I would say) we don't fail but print a rather strange message "Set id of partition /dev/sda1 to (0x06)". With non-meaningful but empty messages we fail. Why is that worse?

Must be me (I'm a bit underpowered today), but the whole meaningfullnes topic does not have a clear meaning to me. :wink:

aschnell: The intention is to not display e.g. "Set id to BIOS Boot (0x102)" since 0x102 is meaningless to the user. 0x102 is a "random" value in libstorage-ng.

On the other hand displaying e.g. "Set id to EFI System Partition (0xEF)" is meaningful since any user can look the standard 0xEF up.

aschnell: Would you prefer ```is_partition_id_value_standardized()```?

ancorgs: Indeed, something like `is_partition_id_standard()` is more obvious to me. Anyway, I'm fine with the current name as long as the explanation in your former comment is anywhere in the code. Your comment is way more explanatory than the inline documentation of `is_partition_id_value_meaningful`.

ancorgs:
LGTM, thanks

PR #177, opened 2016-12-22
- added move constructor and assignment operator

aschnell:
Minor optimization found by coverity.

See http://en.cppreference.com/w/cpp/language/move_constructor and http://en.cppreference.com/w/cpp/language/move_assignment.


shundhammer:
LGTM

PR #180, opened 2017-01-16
renumber logical partitions

aschnell:

### DIFF
--- a/storage/Devices/EncryptionImpl.cc
+++ b/storage/Devices/EncryptionImpl.cc
@@ -336,7 +336,7 @@ namespace storage
     Encryption::Impl::do_rename_etc_crypttab(const Actiongraph::Impl& actiongraph,
 					     const Device* lhs) const
     {
-        // TODO
+        ST_THROW(LogicException("stub do_rename_etc_crypttab called"));
     }
 
 
--- a/storage/Devices/EncryptionImpl.cc
+++ b/storage/Devices/EncryptionImpl.cc
@@ -112,7 +112,7 @@ namespace storage
 
 	actions.push_back(new Action::Create(get_sid()));
 	actions.push_back(new Action::Activate(get_sid()));
-	actions.push_back(new Action::AddEtcCrypttab(get_sid()));
+	actions.push_back(new Action::AddToEtcCrypttab(get_sid()));
 
 	actiongraph.add_chain(actions);
     }
@@ -129,7 +129,7 @@ namespace storage
 	// TODO depends on mount-by, whether there actually is an entry in crypttab
 	if (get_blk_device()->get_name() != lhs.get_blk_device()->get_name())
 	{
-	    actions.push_back(new Action::RenameEtcCrypttab(get_sid()));
+	    actions.push_back(new Action::RenameInEtcCrypttab(get_sid()));
 	}
 
 	actiongraph.add_chain(actions);
@@ -141,7 +141,7 @@ namespace storage
     {
 	vector<Action::Base*> actions;
 
-	actions.push_back(new Action::RemoveEtcCrypttab(get_sid()));
+	actions.push_back(new Action::RemoveFromEtcCrypttab(get_sid()));
 	actions.push_back(new Action::Deactivate(get_sid()));
 	actions.push_back(new Action::Delete(get_sid()));
 
@@ -290,7 +290,7 @@ namespace storage
 
 
     Text
-    Encryption::Impl::do_add_etc_crypttab_text(Tense tense) const
+    Encryption::Impl::do_add_to_etc_crypttab_text(Tense tense) const
     {
 	Text text = tenser(tense,
 			   // TRANSLATORS: displayed before action,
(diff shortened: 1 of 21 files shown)


storage/Devices/MsdosImpl.cc
```diff
@@ -63,6 +63,29 @@ namespace storage
...
+
+	// After deleting a logical partition the numbers of partitions with
+	// higher numbers are shifted.
+	if (old_type == PartitionType::LOGICAL)
+	{
+	    vector<Partition*> partitions = get_partitions();
+            for (Partition* tmp : partitions)
+	    {
+		unsigned int number = tmp->get_number();
+		if (number > old_number)
```
shundhammer: I know that in this context this cannot happen, but I'd feel a lot more comfortable with `if (number != old_number)`. Just to make sure.

storage/Devices/MsdosImpl.cc
```diff
@@ -63,6 +63,29 @@ namespace storage
...
+		    tmp->get_impl().set_number(number - 1);
```
shundhammer: Similarly, I wouldn't rely on that partition's old number; I'd keep my own counter.

storage/Devices/PartitionImpl.cc
```diff
@@ -133,11 +133,29 @@ namespace storage
...
+	return atoi(name.substr(pos + 1).c_str());
+    }
+
+
+    void
+    Partition::Impl::set_number(unsigned int number)
+    {
+	const string& name = get_name();
+
+	string::size_type pos = name.find_last_not_of("0123456789");
```
shundhammer: IMHO that parsing from this and the previous function should be moved out into a separate function that can be tested separately. This looks much too fragile to me to be tested only inside such a relatively complex and specialized environment.

I propose a function to split a partition device name ("/dev/sda4" etc.) into the name part and the number part, i.e. "/dev/sda" and 4 (int). It whould throw exceptions upon a parse error. It might return a `std::pair<string, int>` or use non-const references as parameters. It would be a static (class) function. It might even be useful to the outside world, so it might be part of some StringUtil class.

Then it would be trivial to add sufficient unit tests to make sure it really works under all circumstances. I really don't like sequences like that `find_last_not_of("0123456789")` being repeated all over the code.

shundhammer: For symmetry reasons, there should also be the other way round: A format() function to build a partition device name from its base device name ("/dev/sda") and its number. Yes, it sounds trivial, but it's a matter of centralizing tests and having good unit test coverage.

storage/Devices/EncryptionImpl.cc
```diff
@@ -294,6 +312,35 @@ namespace storage
 
 
     Text
+    Encryption::Impl::do_rename_etc_crypttab_text(const Device* lhs, Tense tense) const
```
shundhammer: That function name implies that the /etc/crypttab file is renamed, but that's not true: Entries **in** that file are renamed, so it should rather be `do_rename_in_etc_crypttab`. Similar for all the functions of this type.

storage/Filesystems/FilesystemImpl.cc
```diff
@@ -783,16 +791,64 @@ namespace storage
 
 
     Text
+    Filesystem::Impl::do_rename_etc_fstab_text(const Device* lhs, const string& mountpoint,
```
shundhammer: Again: Rename file vs. rename **in** file. This should be `do_rename_in_etc_fstab_text()`.

storage/Filesystems/FilesystemImpl.h
```diff
@@ -237,6 +240,21 @@ namespace storage
 	};
 
 
+	class RenameEtcFstab : public Modify
```
shundhammer: Similar naming inssue: This should be `RenameInEtcFstab`.

storage/Devices/MsdosImpl.cc
ancorgs: Do you mean going through all the partitions on every change and not only the affected ones? Not sure what's the point.

storage/Devices/PartitionImpl.cc
ancorgs: I fully agree. I will try to implement this.

storage/Devices/MsdosImpl.cc
shundhammer: There are really two checks here (or rather, there should be two): One for the starting point from where to renumber, and one for the new expected number. Since it's all rolled into one due to the iteration, it might make sense to calculate the new expected number for each one, and if the actual number is different from that one, change it (which might involve more complex things like renaming them in files like /etc/crypttab and /etc/fstab).

Admittedly, if everything has been correct before, the outcome should be the same. But I don't like to rely on things being correct before.

aschnell: I changed the if from ```<``` tp ```!=```.

storage/Devices/EncryptionImpl.cc
aschnell: Will be changed (taken from Ancors branch).

ancorgs:
LGTM

PR #181, opened 2017-01-23
Added Travis support (use Docker for building)

lslezak:
See https://hub.docker.com/r/yastdevel/storage-ng/ and https://github.com/yast/docker-storage-ng for the details about the `yastdevel/storage-ng` Docker image used for building.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+sudo: required
+language: bash
+services:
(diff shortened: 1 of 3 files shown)


ancorgs:
To be honest, I don't understand most of it (docker & travis magic). But I saw it working so LGTM.

PR #182, opened 2017-01-27
Superset of "renumber logical partitions"

ancorgs:
This pull request includes:
 * All the commits from #180 (cherry-picked because @aschnell's master would need a rebase so I cannot "duplicate" the PR cleanly).
 * Two commits to fix issues pointed during code review of #180 (2 out of 3 issues addressed).
 * An extra commit implementing dependencies for deleting partitions (first deletions in descending order and afterwards creations in ascending one). 99% of the code is taken from @aschnell patch at https://trello.com/c/EPvvvAs3/ (I left out a lot of code from that patch for which I didn't get the point).

Dependencies and renumbering of partitions tested manually. Everything seems to work. Unit tests not added.

### DIFF
--- a/storage/Utils/AppUtil.cc
+++ b/storage/Utils/AppUtil.cc
@@ -452,4 +452,28 @@ string afterLast(const string& s, const string& pat )
 
 const string app_ws = " \t\n";
 
+    /**
+     * Breaks a device name like "/dev/sda2" into "/dev/sda" and 2
+     *
+     * Throws an exception if the provided name does not contain the number
+     * part.
+     */
+    std::pair<string, unsigned int>
+    device_to_name_and_number(const string& full_name)
+    {
+	string::size_type pos = full_name.find_last_not_of("0123456789");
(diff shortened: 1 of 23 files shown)


storage/Utils/AppUtil.cc
```diff
@@ -452,4 +452,28 @@ string afterLast(const string& s, const string& pat )
...
+	return std::make_pair(full_name.substr(0, pos + 1), atoi(full_name.substr(pos + 1).c_str()));
+    }
+
+    /**
+     * Inverse of device_to_name_and_number
+     */
+    string
+    name_and_number_to_device(const string& name, unsigned int number)
+    {
+	return (name + to_string(number));
```
aschnell: ```return``` is a statement not a function thus no braces are needed.

aschnell:
Partly included in https://github.com/openSUSE/libstorage-ng/pull/180, rest will follow in another PR.

PR #183, opened 2017-02-08
- add dependencies to actions of children of partition table

aschnell:


storage/Action.cc
```diff
@@ -88,20 +87,23 @@ namespace storage
...
-			vector<Actiongraph::Impl::vertex_descriptor> tmp = actiongraph.actions_with_sid(child_sid, ONLY_LAST);
-			if (!tmp.empty())
+			Devicegraph::Impl::adjacency_iterator vi2, vi2_end;
+			for (boost::tie(vi2, vi2_end) = adjacent_vertices(q, actiongraph.get_devicegraph(LHS)->get_impl().graph); vi2 != vi2_end; ++vi2)
 			{
-			    // Make sure it's a delete action
-			    const Action::Base* tmp_action = actiongraph[tmp.front()];
-			    if (dynamic_cast<const Action::Delete*>(tmp_action))
-				actiongraph.add_edge(tmp.front(), vertex);
+			    sid_t child_sid = actiongraph.get_devicegraph(LHS)->get_impl()[*vi2]->get_sid();
```
shundhammer: This complex expression is very hard to read. How about breaking it up and using some more variables to make clear what happens?

storage/Action.cc
```diff
@@ -88,20 +87,23 @@ namespace storage
...
+
+			    vector<Actiongraph::Impl::vertex_descriptor> tmp = actiongraph.actions_with_sid(child_sid, ONLY_LAST);
+			    if (!tmp.empty())
+			    {
+				// Make sure it's a delete action
+				const Action::Base* tmp_action = actiongraph[tmp.front()];
+				if (dynamic_cast<const Action::Delete*>(tmp_action))
```
shundhammer: If this is used more often, this might be a good candidate for a one-line inline function isDeleteAction().

aschnell: The code was only indented differently. But there is already a is_delete() functions and I can use it here.

shundhammer:
Just that one minor readability issue left over. Other than that, LGTM.

storage/Action.cc
aschnell: The code was only indented differently. Otherwise I agee, the code should be better readable.

PR #184, opened 2017-02-09
- implemented do_rename_in_etc_crypttab()

aschnell:

### DIFF
--- a/storage/Devices/LuksImpl.cc
+++ b/storage/Devices/LuksImpl.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 SUSE LLC
+ * Copyright (c) [2016-2017] SUSE LLC
  *
  * All Rights Reserved.
  *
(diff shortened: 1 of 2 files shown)


storage/Devices/LuksImpl.cc
```diff
@@ -291,9 +291,25 @@ namespace storage
...
+
+	EtcFstab fstab(storage.get_impl().prepend_rootprefix("/etc"));	// TODO pass as parameter
+
+	FstabKey key(luks_lhs->get_blk_device()->get_name(), "");
+	FstabChange entry;
+	entry.device = get_blk_device()->get_name();
+	entry.dentry = get_name();
+	entry.encr = EncryptionType::LUKS;
+
+	fstab.removeEntry(key);
```
joseivanlopez: Just an observation. Maybe, it would be more clear to use the same object type for both, `EtcFstab#removeEntry` and `EtcFstab#addEntry`. That is, use key or entry in both methods:
```
fstab.removeEntry(entry);
fstab.addEntry(entry);
```
or
```
fstab.removeEntry(key);
fstab.addEntry(key, entry);
```
Another option could be to add a method `EtcFstab#removeEntryByKey`.

testsuite/fstab.cc
```diff
@@ -355,6 +355,38 @@ BOOST_AUTO_TEST_CASE(update6)
...
+    fstab.addEntry(entry);
+    fstab.flush();
+
+    check({
+	""
+    }, {
+	"cr_test         /dev/sdb5            none       none"
+    });
+}
+
```
joseivanlopez: Code of this test is very similar to `EtcFstab#removeEntry`. Is it not possible call the method?

aschnell: There is a updateEntry() function but it does not work. As I wrote the best would be to rewrite the EtcFstab class (and maybe split EtcCrypttab).

joseivanlopez:
LGTM

PR #187, opened 2017-02-15
Bind exceptions for BlkDevice::get_encryption

joseivanlopez:
Exceptions of  `BlkDevice::get_encryption` do not propagate to ruby.

Add entries to storage-catches to bind them.


aschnell:
LGTM

PR #191, opened 2017-02-20
- split off class BlkFilesystem from class Filesystem

aschnell:
So far all supported filesystems are block device based filesystems. For NFS and tmpfs (if support is added) many functions do not apply. So split off the new class BlkFilesystem from class Filesystem.

Patch for yast2-storage-ng is in preparation.


wfeldt:
lgtm

PR #193, opened 2017-02-24
- redirect info calls to probed devicegraph

aschnell:
Not for a specific PBI.


aschnell:
Will be replace by improved pull request.

PR #196, opened 2017-02-28
- interface improvements

aschnell:
Not for a specific PBI.

### DIFF
--- a/storage/Filesystems/BlkFilesystem.h
+++ b/storage/Filesystems/BlkFilesystem.h
@@ -46,13 +46,13 @@ namespace storage
 	static std::vector<BlkFilesystem*> get_all(Devicegraph* devicegraph);
 	static std::vector<const BlkFilesystem*> get_all(const Devicegraph* devicegraph);
 
-	virtual bool supports_label() const = 0;
-	virtual unsigned int max_labelsize() const = 0;
+	bool supports_label() const;
+	unsigned int max_labelsize() const;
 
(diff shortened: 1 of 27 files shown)


storage/Filesystems/BlkFilesystem.h
```diff
@@ -46,13 +46,13 @@ namespace storage
 	static std::vector<BlkFilesystem*> get_all(Devicegraph* devicegraph);
 	static std::vector<const BlkFilesystem*> get_all(const Devicegraph* devicegraph);
 
-	virtual bool supports_label() const = 0;
-	virtual unsigned int max_labelsize() const = 0;
+	bool supports_label() const;
```
shundhammer: Is it really a good idea to move the pure virtual functions from the public class to the impl class?
If you try to call such a pure virtual function on the base class, the linker will complain. But what happens when this is now in the impl class? Will the linker still detect the problem?

shundhammer:
LGTM with one open question (the pure virtuals)

storage/Filesystems/BlkFilesystem.h
aschnell: BlkFilesystem is still an abstract class (since clone in Device is not defined). So you cannot call those functions on the base class (BlkFilesystem) but only the concrete derived classes (e.g. Ext4, Btrfs).

shundhammer: OK.

PR #198, opened 2017-03-06
Imported https://github.com/shundhammer/commented-config-file

shundhammer:
See also

https://github.com/shundhammer/commented-config-file

### DIFF
--- a/storage/Utils/CommentedConfigFile.cc
+++ b/storage/Utils/CommentedConfigFile.cc
@@ -1,0 +1,364 @@
+/*
+ * File: CommentedConfigFile.cc
+ *
+ * Copyright (c) 2017 Stefan Hundhammer <Stefan.Hundhammer@gmx.de>
+ *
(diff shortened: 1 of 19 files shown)


storage/Utils/CommentedConfigFile.cc
```diff
@@ -0,0 +1,366 @@
...
+        content_ret = line.substr( 0, pos-1 );
+        comment_ret = line.substr( pos );
+    }
+
+    strip_trailing_whitespace( content_ret );
+}
+
+
+void CommentedConfigFile::strip_trailing_whitespace( string & line )
+{
```
aschnell: Maybe you could use http://www.boost.org/doc/libs/1_56_0/doc/html/boost/algorithm/trim_right_copy.html here.

shundhammer: Yes, I looked at that, and the resulting code was a lot less clear than this, so I preferred this own implementation.

PR #200, opened 2017-03-09
- split off Mountable from Filesystem

aschnell:
For https://trello.com/c/yES1u2VG/425-8-libstorageng-btrfs-subvolumes.

### DIFF
--- a/storage/Filesystems/MountableImpl.h
+++ b/storage/Filesystems/MountableImpl.h
@@ -1,0 +1,214 @@
+/*
+ * Copyright (c) [2014-2015] Novell, Inc.
+ * Copyright (c) [2016-2017] SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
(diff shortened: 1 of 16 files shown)


storage/Filesystems/MountableImpl.h
```diff
@@ -0,0 +1,214 @@
...
+
+	    virtual Text text(const Actiongraph::Impl& actiongraph, Tense tense) const override;
+	    virtual void commit(const Actiongraph::Impl& actiongraph) const override;
+
+	    const string mountpoint;
+
+	};
+
+    }
+
```
joseivanlopez: The interface specification of these 4 classes looks like quite similar. Maybe it could be written only once, and then each subclass offers its own implementation. But I do not sure if it is worth.

```
class MountableAction : public Modify
{
public:

    MountableAction(sid_t sid, const string& mountpoint)
    : Modify(sid), mountpoint(mountpoint) {}

    virtual Text text(const Actiongraph::Impl& actiongraph, Tense tense) const override;
    virtual void commit(const Actiongraph::Impl& actiongraph) const override;

    const string mountpoint;
};

class Mount : public MountableAction { }

class Umount : public MountableAction { }

class AddToEtcFstab : public MountableAction
{ 
    virtual void add_dependencies(Actiongraph::Impl::vertex_descriptor vertex,
                                  Actiongraph::Impl& actiongraph) const override;
}

class RemoveFromEtcFstab : public MountableAction { }

aschnell: I'm not sure I understand the issue. Of course the interfaces look identical since they are all derived from Action::Modify. But when you reimplement a function in C++ you have to provide the declaration in the class definition so in this case even when deriving from MountableAction the text and commit functions must be declared in Mount, Umount, ...

joseivanlopez: Right, I did not realize about that detail. Then, I have not said anything due to the constructor is the uniq method with the same implementation in the four classes.

PR #206, opened 2017-03-15
New EtcFstab

shundhammer:
https://trello.com/c/iXbCO3ru/517-8-libstorage-ng-redesign-etcfstab


testsuite/probe/bcache1-devicegraph.xml
```diff
@@ -166,14 +166,14 @@
       <uuid>6990d6be-8fa8-4be7-8cf3-66ad1448864e</uuid>
       <mountpoint>swap</mountpoint>
       <mount-by>uuid</mount-by>
-      <fstab-options>defaults</fstab-options>
+      <mount-opts>defaults</mount-opts>
```
aschnell: Would it not be more consistent to remove the "defaults" since it is also not present in the internal data structures?

storage/EtcCrypttab.h
```diff
@@ -0,0 +1,182 @@
...
+
+	string	  crypt_device; // Resulting device below /dev/mapper
+	string	  block_device; // Underlying filesystem etc.; "UUID=" supported
+	string	  password;	// optional; "none" or "-" for "ask"
+	CryptOpts crypt_opts;	// optional
+    };
+
+
+    /**
+     * Class representing /etc/fstab.
```
aschnell: representing /etc/crypttab

storage/EtcCrypttab.h
```diff
@@ -0,0 +1,182 @@
...
+
+	/**
+	 * Add an entry. This is just an alias for append() to maintain
+	 * symmetry with the EtcFstab class.
+	 **/
+	void add( CrypttabEntry * entry ) { append( entry ); }
+
+	/**
+	 * Return entry no. 'index' or 0 if 'index' is out of range.
+	 *
```
aschnell: I would prefer an exception if index is out of range.

storage/Utils/ColumnConfigFile.h
```diff
@@ -155,6 +170,14 @@ class ColumnConfigFile: public CommentedConfigFile
      **/
     void set_max_column_width( int new_size );
 
+    /**
+     * Return entry no. 'index' or 0 if 'index' is out of range.
+     *
+     * This is a covariant of the (non-virtual) base class method to reduce
+     * the number of dynamic_casts.
+     **/
+    ColumnConfigFile::Entry * get_entry( int index ) const;
```
aschnell: Add '''override'''. Maybe also at other places.

storage/EtcFstab.cc
```diff
@@ -19,1000 +20,424 @@
...
 
 
 namespace storage
 {
-    using namespace std;
-
 
-    string
-    getExtDataMode(const string& fstab_options)
+    MountOpts::MountOpts( const string & opt_string )
```
aschnell: Empty constructor looks wrong.

testsuite/probe/bcache1-devicegraph.xml
shundhammer: Right. It is filtered out in the next step anyway, but there is no good reason to keep this here in the XML files.

Removed here and in 2 other XML files.

storage/EtcCrypttab.h
shundhammer: Fixed.

storage/EtcFstab.cc
shundhammer: Fixed.

storage/Filesystems/MountableImpl.cc
```diff
@@ -242,19 +250,27 @@ namespace storage
...
-	EtcFstab fstab(storage.get_impl().prepend_rootprefix("/etc"));	// TODO pass as parameter
+	EtcFstab fstab;
+        fstab.read(storage.get_impl().prepend_rootprefix("/etc"));	// TODO pass as parameter
+
+        FstabEntry * entry = new FstabEntry();
+        entry->set_device(get_mount_by_name());
+        entry->set_mount_point(mountpoint);
+        entry->set_mount_opts(get_mount_opts());
+
+        // FIXME: Why is this not a FsType in the first place?
```
aschnell: Likely because the old EtcFstab class expected a string. I would propose to move the FsType enum from Filesystem to Mountable.

storage/SystemInfo/ProcMounts.cc
```diff
@@ -37,48 +41,69 @@ namespace storage
...
+
+            if ( entry )
+            {
+                const string & device = entry->get_device();
+
+                if ( entry->get_fs_type() == FsType::UNKNOWN ||
+                     device == "rootfs" ||
+                     device == "/dev/root" )
+                {
+                    // Get rid of all the usesless stuff that clutters /proc/mounts
```
aschnell: 'useless'

shundhammer: Fixed.

It's important to keep our rants in order. 😄

storage/Utils/ColumnConfigFile.cc
```diff
@@ -65,8 +71,10 @@ string ColumnConfigFile::Entry::format()
 }
 
 
-bool ColumnConfigFile::Entry::parse( const string & line )
+bool ColumnConfigFile::Entry::parse( const string & line, int line_no )
 {
+    (void) line_no;
```
aschnell: This can also be done with ```int __attribute__((__unused__)) line_no``` in the signature. We might define a macro for that and then enable the compiler warning.

shundhammer: Good idea.

storage/Utils/CommentedConfigFile.h
```diff
@@ -294,7 +297,7 @@ class CommentedConfigFile: private boost::noncopyable
     /**
      * Return entry no. 'index' or 0 if 'index' is out of range.
      **/
-    Entry * get_entry( int index );
+    Entry * get_entry( int index ) const;
```
aschnell: Shouldn't the return type be ```const Entry*``` when the function is ```const```.

shundhammer: That's a philosophical question... this 'const' promises it won't modify the CommentedConfigFile object. That's not quite the same as promising not to modify one of the objects it keeps pointers to. That's how large libs like Qt do it all the time; getting a pointer to some items is a const operation, even though the item itself may be modified by a non-const pointer.

If you see it that strictly, there are almost no real const methods left because this non-constness here will spread out like a disease.

storage/Filesystems/MountableImpl.h
```diff
@@ -57,8 +58,16 @@ namespace storage
 	MountByType get_mount_by() const { return mount_by; }
 	void set_mount_by(MountByType mount_by);
 
-	const list<string>& get_fstab_options() const { return fstab_options; }
-	void set_fstab_options(const list<string>& fstab_options);
+	const MountOpts & get_mount_opts() const { return mount_opts; }
```
aschnell: Without swig parsing the header with MountOpts (and someday also CryptOpts) this will not work in the generated bindings. I propose to simply use ```vector<string>``` in the external API.

aschnell:
LGTM

ancorgs:
@shundhammer, @aschnell It looks to me like this PR (or any other of the very recent ones) broke the StorageNG ISO. Look how much red we have here: 
https://yast-openqa.suse.cz/tests/overview?distri=opensuse&version=Tumbleweed&build=388.16&groupid=2

The problem is that during installation right after creating the filesystems and mounting them in `/mnt/`, yast-packager expects `/mnt/etc` to not be there (it's a brand new filesystem) but now it is.

Not sure if that's reasonable/intentional or not. Was that the intention?

Anyway, no matter if we "fix" libstorage-ng to act as the old one, in my opinion we should fix yast-packager to be more robust and not fail if `/mnt/etc` is there.

aschnell:
Maybe first fix https://build.opensuse.org/project/monitor/YaST:storage-ng bu merging https://github.com/openSUSE/libstorage-ng/pull/209.

ancorgs:
Closer inspection revealed the problem is not the presence of `/mnt/etc` (that would not make YaST crash). The problem is that `/mnt/etc/` is not a directory, but a file (with the content that should be in `/mnt/etc/fstab`).

PR #207, opened 2017-03-16
- added example program

aschnell:


examples/transmogrify.cc
```diff
@@ -0,0 +1,80 @@
...
+	}
+	// remove device with sid and its descendants
+	else if (command == "rm+")
+        {
+	    Device* device = staging->find_device(std::stoi(commands[index++]));
+	    device->remove_descendants();
+	    staging->remove_device(device);
+        }
+	// only remove descendants of device with sid
+	else if (command == "rm+-")
```
shundhammer: Those command names are clearly too speaking for a decent BrainFuck implementation. I think we can do better than that. 😃

shundhammer:
LGTM

examples/transmogrify.cc
aschnell: We will see when more commands are added.

PR #211, opened 2017-03-17
- fixed mockup recording for /etc/fstab

aschnell:


aschnell:
StorageNG-Test-Build389.2-Media.iso is available meanwhile.

PR #214, opened 2017-03-20
Handle filesystem type 'unknown' better in EtcFstab

shundhammer:
Since the EtcFstab uses the FsType enum for all operations, this is limited to the filesystem types known to that enum. But that does not include any new ones or even those for all the kernel-only filesystems like "sysfs", "proc", "cgroup" etc.

This change uses one deeper layer of the inheritance hierarchy of EtcFstab: The ColumnConfigFile class still has all the fields of each entry line as string. So for fs type "unknown", the EtcFstab now tries to use its parent column[3] if it has any content. For new entries, it has to write something in that field to maintain the syntax of /etc/fstab which alway expects that field, so it uses "unknown" (and writes an error to the log).

### DIFF
--- a/storage/EtcFstab.cc
+++ b/storage/EtcFstab.cc
@@ -174,7 +174,20 @@ namespace storage
 	int col = 0;
 	set_column( col++, EtcFstab::fstab_encode( device      ) );
 	set_column( col++, EtcFstab::fstab_encode( mount_point ) );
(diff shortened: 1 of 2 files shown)


jsrain:
LGTM if the pending check passes

shundhammer:
`make check` works locally. Travis is very slow today.

PR #216, opened 2017-03-20
Generic diff tool in CommentedConfigFile and now logging diffs in EtcFstab

shundhammer:


shundhammer:
Disregard this PR; I'll move that diff stuff out into a separate class for clarity.

PR #219, opened 2017-03-22
Diff for EtcFstab

shundhammer:
This adds a generic diff class for string vectors (taken from https://github.com/shundhammer/commented-config-file) and uses it to log only the diff of each EtcFstab change, not the whole file.

There are no unit tests for the Diff class yet, but I tested it extensively manually in the upstream project: The `diff -u` command is the reference what it should do.


wfeldt:
lgtm

PR #226, opened 2017-03-29
- work on MD size calculation

aschnell:
For https://trello.com/c/EMKFnYiV/275-8-libstorage-ng-md-arrays.

### DIFF
--- a/storage/Devices/MdImpl.cc
+++ b/storage/Devices/MdImpl.cc
@@ -438,6 +438,12 @@ namespace storage
     void
     Md::Impl::calculate_region_and_topology()
     {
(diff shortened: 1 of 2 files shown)


storage/Devices/MdImpl.cc
```diff
@@ -438,6 +438,12 @@ namespace storage
     void
     Md::Impl::calculate_region_and_topology()
     {
+	// Calculating the exact size of a MD is difficult. Since a size to
+	// big can lead to severe problems later on, e.g. a partition not
```
joseivanlopez: Since a size **too** big, right?

aschnell: Indeed.

PR #227, opened 2017-03-29
- work on MD support

aschnell:
For https://trello.com/c/EMKFnYiV/275-8-libstorage-ng-md-arrays.

### DIFF
--- a/storage/Filesystems/BtrfsImpl.cc
+++ b/storage/Filesystems/BtrfsImpl.cc
@@ -78,6 +78,32 @@ namespace storage
     }
 
 
+    BtrfsSubvolume*
+    Btrfs::Impl::get_default_btrfs_subvolume()
+    {
+	for (BtrfsSubvolume* btrfs_subvolume : get_btrfs_subvolumes())
+	{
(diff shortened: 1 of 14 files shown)


storage/Filesystems/BtrfsImpl.cc
```diff
@@ -78,6 +78,32 @@ namespace storage
...
+    {
+	for (const BtrfsSubvolume* btrfs_subvolume : get_btrfs_subvolumes())
+	{
+	    if (btrfs_subvolume->is_default_btrfs_subvolume())
+		return btrfs_subvolume;
+	}
+
+	ST_THROW(Exception("no default btrfs subvolume found"));
+    }
+
```
joseivanlopez: Is it possible to avoid to write exactly the same implementation twice? Maybe using `const_cast` ?

aschnell: In some cases like here this could be done with a const_cast. Some people like it and others do not like it.

For more complex return values it is not easy, e.g. it is not possible to const_cast from ```vector<const type*>``` to ```vector<type*>```.

PR #230, opened 2017-04-03
- coding style and unit test

aschnell:
Not for a specific PBI.

### DIFF
--- a/storage/SystemInfo/CmdDmraid.cc
+++ b/storage/SystemInfo/CmdDmraid.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) [2004-2014] Novell, Inc.
+ * Copyright (c) 2017 SUSE LLC
  *
  * All Rights Reserved.
  *
@@ -58,7 +59,7 @@ namespace storage
 		list<string>::const_iterator ci = sl.begin();
 		string name = *ci;
 		advance(ci, 3);
-		entry.raidtype = *ci;
+		entry.raid_type = *ci;
 
 		while (it != lines.end() && boost::starts_with(*it, "/dev/"))
 		{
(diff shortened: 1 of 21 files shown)


storage/SystemInfo/CmdDmraid.cc
```diff
@@ -80,22 +81,22 @@ namespace storage
...
+	vector<string> ret;
 	for (const_iterator it = data.begin(); it != data.end(); ++it)
 	    ret.push_back(it->first);
 	return ret;
     }
 
 
     bool
-    CmdDmraid::getEntry(const string& name, Entry& entry) const
+    CmdDmraid::get_entry(const string& name, Entry& entry) const
```
shundhammer: Hm... IMHO a `get_xy()` function should return the xy as its return value, not a boolean ok and as a byproduct modify a non-const reference.

Alternative approaches:

- `Entry CmdDmraid::get_entry( const string & name ) const
  that throws an exception if not found - this would probably be appropriate in this case since not getting a valid result for a name (that you fetched from some other place here) is most likely a programmatic error.

- `Entry * get_entry( const string & name ) const
  that returns a null pointer if not found.

But always returning the less important ok flag and having to do mental acrobatics for that is not the best solution IMHO.

shundhammer:
If you decide to not go along with my suggestion, feel free to merge anyway.

The rest is LGTM.

storage/SystemInfo/CmdDmraid.cc
aschnell: I agree that the function interface is not nice. I will change it to return ```const Entry&``` for all currently used ```get_entry()``` functions in SystemInfo.

shundhammer:
LGTM

PR #235, opened 2017-04-06
- check types to avoid exception

aschnell:
Found while working on https://trello.com/c/WAGRJLgv/542-5-libstorage-ng-support-of-alternative-device-naming-schemes.


ancorgs:
LGTM

PR #238, opened 2017-04-20
- moved mount point data to dedicated class in devicegraph

aschnell:
Move mount point data to dedicated MountPoint class in devicegraph. This 1) make the interface consistent, 2) simplyfies many functions creating actions and 3) makes the mount points visible in the device graph pictures. Also use the default mount-by method when creating a MountPoint (see documentation).

For https://trello.com/c/WAGRJLgv/542-5-libstorage-ng-support-of-alternative-device-naming-schemes.


ancorgs:
I have not dived into the code (too much C++ for me 😉), but somehow I dislike the fact that the new Mountpoint class inherits from Device and is represented in the devicegraph. So far, everything descending from Device is actually... well... a device (i.e. something you can in most cases find in the `/dev` directory[1]). That's not the case with Mountpoint. Making the new class a subclass of Device looks too artificial to me and having the mountpoints as devicegraph nodes is unexpected to me.

[1] With some reasonable exceptions like PartitionTable, that you probably cannot find in `/dev` but that makes a lot of sense as subclass of Device and as a devicegraph node.

aschnell:
Filesystems and btrfs subvolumes are also not devices in the normal sense. Neither are LVM physical volumes and partition tables as you already mentioned.

If you have a better name for the Device class please tell so (but please nothing like StorageObject).

ancorgs:
I actually think Device is a good name. Or at least it was until the introduction of Mountpoint. Having Filesystem, BtrfsSubvolume, LvmPv or PartitionTable as devices simply makes sense to me (even if they don't have counterpart in `/dev`[1]). Conceptually and intuitively I see strong reasons for having them in the devicegraph.

But, to me, the mountpoints (as long as I agree they deserve their own class) are just a regular property of the filesystems. Just like the partition_slots are a property of the partition tables, or the resize_infos are a property of the block devices. All of them deserve a dedicated class, but none of them deserve to be called a device and to be included in the devicegraph. IMHO.

Anyway, I will try to come with something more solid than "conceptually and intuitively" if you disagree. :wink:

[1] It was probably misleading from my part to mention `/dev` as a criteria to consider something a device.

storage/Filesystems/MountPointImpl.cc
```diff
@@ -0,0 +1,492 @@
...
+    MountPoint::Impl::add_modify_actions(Actiongraph::Impl& actiongraph, const Device* lhs_base) const
+    {
+	Device::Impl::add_modify_actions(actiongraph, lhs_base);
+
+	const Impl& lhs = dynamic_cast<const Impl&>(lhs_base->get_impl());
+
+	if (!lhs.active && active)
+	{
+	    Action::Base* action = new Action::Mount(get_sid());
+	    actiongraph.add_vertex(action);
```
aschnell: A vector and using add_chain also adds dependencies between the actions of the vector. So just changing the code here will result in a different actiongraph.

aschnell:
Did you ever look at the devicegraph the old expert partitioner shows? The mount points are also dedicated objects there and nobody ever complained in ten years about that.

I would also not call mount points a property of the filesystem but a property of the storage stack.

PR #246, opened 2017-04-28
- removed probing in constructor and added probe function

aschnell:
For https://trello.com/c/li8xjUAE/538-5-storageng-probe-inactive-luks.


aschnell:
Thanks. I will update misc/activate.py.

PR #259, opened 2017-05-11
Compound actions for simplified summary

joseivanlopez:
This is a very very premature PR only for feedback purposes.

### Summary

The main idea is to have a `CompoundAction` class that basically represents a collection of *commit actions* sharing a `target_device`. This class is the responsible of generating the text representation of a set of *commit actions*.

Then we have another class, `CompoundActionCreator`. This class generates the corresponding set of `CompoundAction` for an *actiongraph*.  The class `Actiongraph` should have an attribute of this type.

Take into account that with this strategy, each subvolume will have its own `CompoundAction` with all the *commit actions* for that specific subvolume. The client is responsible to order and filter the compound actions properly. The library API will have all necessary methods for that.  For example, in the ruby part we will do something like that:

```
compund_actions = actiongraph.compound_actions

delete_actions = compound_actions.select(&:delete_action?)
root_action = compound_actions.detect(&:action_for_root?)
home_action = compound_actions.detect(&:action_for_home?)
subvolumes = compound_actions.select {|a| a.action_for_subvolume?}
```

The API needs to be discussed, of course.

### State

* Commit actions are grouped by *target device* but delete actions are not generated as separated `CompoundAction`.
* The text of a `CompoundAction` is simply the joining of the text for each *commit actions*. It is necessary to generate a specific text depending on the set of *commit actions*.
* `CompoundActionCreator` is not added to `Actiongraph` yet.

Regarding to the namespace, I am not sure if it is better to populate the current `Action` namespace, so we will have:

```
namespace Action
{
    class Compound {};

    class Compound::Creator {};
}

Action::Compound::Creator creator; // instead of as now it is: ActionCompoundCreator
```


storage/CompoundActionCreator.h
```diff
@@ -0,0 +1,74 @@
...
+    {
+
+    public:
+
+	CompoundActionCreator(const Actiongraph* actiongraph);
+	~CompoundActionCreator();
+
+	std::vector<const Action::Base*> get_commit_actions() const;
+
+	std::vector<CompoundAction*> get_compound_actions() const;
```
aschnell: This should return ```vector<const CompoundAction*>```.

storage/CompoundActionCreator.h
```diff
@@ -0,0 +1,74 @@
...
+
+	void group_commit_actions();
```
aschnell: Why not make this function private and call it in the constructor?

storage/CompoundActionCreatorImpl.h
```diff
@@ -0,0 +1,83 @@
...
+	
+	const Device* target_device_for_encryption(const Device* device) const;
+
+	const Device* target_device_for_mount_point(const Device* device) const;
+
+    private:
+
+	const Actiongraph* actiongraph;
+    
+	vector<CompoundAction*> compound_actions;
```
aschnell: Please use something without manual resource handling.

aschnell:
The overall idea seems reasonable to me. Stefan should also have a look at it.

I would prefer to reserve the Action namespace for the "old small" actions of the actiongraph.

.gitignore
```diff
@@ -21,3 +21,4 @@ cov-int
 /.obsdir/
 /obs-package-from-git
 compile
+*.swp
```
aschnell: Move such editor specific rules to ~/.gitignore.

storage/CompoundAction/BtrfsFormater.cc
```diff
@@ -0,0 +1,147 @@
...
+#include "storage/Filesystems/MountPoint.h"
+
+
+namespace storage
+{
+
+    BtrfsFormater::BtrfsFormater(const CompoundAction::Impl* compound_action)
+    : CompoundActionFormater(compound_action) 
+    {
+	this->btrfs = to_btrfs(compound_action->get_target_device());
```
aschnell: Move to member initializer lists above.

storage/CompoundAction/BtrfsFormater.cc
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+
+    BtrfsFormater::~BtrfsFormater() {}
```
aschnell: No need to define and implement the default destructor.

storage/CompoundAction/BtrfsFormater.cc
```diff
@@ -0,0 +1,147 @@
...
+	    return default_text();
+    }
+
+    
+    Text
+    BtrfsFormater::delete_text() const
+    {
+        Text text = tenser(tense,
+                           _("Delete filesystem %1$s on %2$s"),
+                           _("Deleting filesystem %1$s on %2$s"));
```
aschnell: I don't think the continuous form is needed. Where and when should that be displayed?

storage/CompoundAction/BtrfsFormater.cc
```diff
@@ -0,0 +1,147 @@
...
+	else
+	    return default_text();
+    }
+
+    
+    Text
+    BtrfsFormater::delete_text() const
+    {
+        Text text = tenser(tense,
+                           _("Delete filesystem %1$s on %2$s"),
```
aschnell: Comments for translators are missing.

storage/CompoundAction/CompoundActionCreatorImpl.h
```diff
@@ -0,0 +1,88 @@
...
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+
+#ifndef STORAGE_COMPOUND_ACTION_CREATOR_IMPL_H
+#define STORAGE_COMPOUND_ACTION_CREATOR_IMPL_H
+
+
+#include <iterator>
```
aschnell: Why is this include needed?

storage/CompoundAction/CompoundActionCreatorImpl.h
```diff
@@ -0,0 +1,88 @@
...
+    public:
+
+	using const_iterator = vector<CompoundAction*>::const_iterator;
+
+	Impl(const Actiongraph* actiongraph);
+	~Impl();
+
+	vector<const Action::Base*> get_commit_actions() const;
+
+	vector<CompoundAction*> get_compound_actions() const;
```
aschnell: Return ```const vector<CompoundAction*>&```?

aschnell:
Unit tests are missing.

storage/CompoundAction/BtrfsFormater.cc
```diff
@@ -0,0 +1,147 @@
...
+ * with this program; if not, contact SUSE LLC.
+ *
+ * To contact SUSE LLC about this file by physical or electronic mail, you may
+ * find current contact information at www.suse.com.
+ */
+
+
+#include <boost/algorithm/string/join.hpp>
+
+#include "storage/CompoundAction/BtrfsFormater.h"
```
aschnell: "Formatter" is the correct spelling.

storage/ActiongraphImpl.cc
```diff
@@ -485,6 +486,28 @@ namespace storage
 
 
     void
+    Actiongraph::Impl::generate_compound_actions(const Actiongraph* actiongraph)
+    {
+	Impl::compound_actions.clear();
```
aschnell: The ```Impl::``` should not be needed.

testsuite/compound-action.cc
```diff
@@ -0,0 +1,121 @@
...
+
+BOOST_AUTO_TEST_CASE(test_size_with_commit_actions)
+{
+    set_staging();
+
+    auto actiongraph = storage->calculate_actiongraph();
+
+    BOOST_CHECK(actiongraph->get_commit_actions().size() == 182);
+
+    BOOST_CHECK(actiongraph->get_compound_actions().size() == 74);
```
aschnell: Checking the number of actions and compound actions is not really useful to check the functionality. Please write small tests that check the resulting strings of the compound actions.

storage/CompoundAction/Formatter.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "storage/CompoundAction/Formatter.h"
+
+
+namespace storage
+{
+
+    using std::string;
+
+    CompoundAction::Formatter::Formatter(const CompoundAction::Impl* compound_action)
+    : compound_action(compound_action), tense(Tense::SIMPLE_PRESENT) {}
```
aschnell: I still don't see why any tense other than simple present should be used. Removing the variable would save lots of code and messages to translate.

testsuite/compound-action.cc
joseivanlopez: Yes, I agree. I will create more specific tests.

storage/CompoundAction/Formatter.cc
joseivanlopez: Ok, I did that to mimic the commit actions. But I will only use present simple if other tense is not necessary for compound actions.

storage/ActiongraphImpl.cc
```diff
@@ -485,6 +486,28 @@ namespace storage
...
+	for (auto action : compound_actions)
+	    Impl::compound_actions.push_back(shared_ptr<CompoundAction>(action));
+    }
+
+
+    vector<const CompoundAction*>
+    Actiongraph::Impl::get_compound_actions() const
+    {
+	vector<const CompoundAction*> actions;
+	for (auto action : compound_actions)
```
aschnell: The variable names are a bit inconsistent: E.g. ```action``` should be named ```compound_action```. The return value could be named ```ret``` to solve the name collision.

storage/CompoundAction/Formatter/Btrfs.cc
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    
+    Text
+    CompoundAction::Formatter::Btrfs::delete_text() const
+    {
+	// TRANSLATORS:
+	// %1$s is replaced by filesystem name (e.g. btrfs),
+	// %2$s is replaced by name of devices separated by
+	// comma (e.g. /dev/sda1, /dev/sda)
```
aschnell: No need to split such short lines.

aschnell: BTW: Using sda and sda1 for btrfs looks impossible.

storage/CompoundAction/Formatter/Btrfs.cc
```diff
@@ -0,0 +1,157 @@
...
+        Text text = _("Delete filesystem %1$s on %2$s");
```
aschnell: Shouldn't the message include the filesystem type?

aschnell: I don't understand why the filesystem type is a variable here although the file is called Formatter/Btrfs.cc.

storage/CompoundAction/Formatter/Partition.cc
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+
+    Text
+    CompoundAction::Formatter::Partition::create_encrypted_pv_text() const
+    {
+	// TRANSLATORS:
+	// %1$s is replaced by partition name (e.g. /dev/sda1),
+	// %2$s is replaced by size (e.g. 2GiB)
+	Text text = _("Create encrypted partition %1$s (%2$s) as LVM physical device");
```
aschnell: Those are called "physical volume".

storage/CompoundAction/Formatter/Partition.cc
```diff
@@ -0,0 +1,413 @@
...
+		       filesystem->get_displayname().c_str());
+    }
+
+
+    Text
+    CompoundAction::Formatter::Partition::fs_text() const
+    {
+	auto filesystem = get_created_filesystem();
+
+	Text text = _("Format partition %1$s (%2$s) with %3$s");
```
aschnell: Move below ```// TRANSLATORS:```.

aschnell:
Is it also possible to have unit tests that show the complete result of e.g. a standard installation? In the end that is what is interesting.

joseivanlopez:
@aschnell a question about your last comment. I think the idea was small test, checking that a set of commit actions results in a specific compound action, and also that this compound action generates the expected sentence. A more general test for checking the result of a standard installation maybe has more sense in the proposal?

aschnell:
@joseivanlopez I mainly didn't like the old test because it just tested the number of actions and compound actions. Those numbers are basically not comprehensible so the test was unmaintanable.

The simple tests now are good but bigger tests that checks e.g. the ordering of compound actions looks also important for me.

testsuite/Makefile.am
```diff
@@ -3,7 +3,7 @@
 #
 
 SUBDIRS = helpers . Utils SystemInfo probe dependencies sorting freeinfo 	\
-	Devices partitions performance commented-config-file
+	  Devices partitions performance commented-config-file CompoundAction
```
aschnell: Please remove new spaces at beginning.

aschnell:
LGTM

PR #267, opened 2017-05-29
[RFC] count coverage in travis

jreidinger:
currently just testing if it works. With switch to docker, it should be easy to count coverage without any special hacks beside defining CXXFLAGS=-coverage to get coverage report.

### DIFF
--- a/.docker.sh
+++ b/.docker.sh
@@ -1,0 +1,6 @@
+#! /bin/bash
+
+set -x
+
+env
+storage-ng-travis-cpp && bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ before_install:
 script:
   # the "storage-ng-travis-cpp" script is included in the base yastdevel/storage-ng image
   # see https://github.com/yast/docker-storage-ng/blob/master/storage-ng-travis-cpp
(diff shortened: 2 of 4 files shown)


.docker.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+
+set -x
+
+env
+storage-ng-travis-cpp && bash <(curl -s https://codecov.io/bash)
```
aschnell: To me it looks very dangerous to automatically execute code downloaded from an untrusted site.

jreidinger: that is exactly reason why we use travis and docker inside. There are no permissions there, nothing to attack. That is why I use it. Even if code gets complete root permission, penetrate docker and even penetrate KVM to control server. It will be travis server and not our.

ancorgs: So apparently we need some extra information on the header of `.docker.sh` explaining what is it used for.

jreidinger: yeah, I even plan to add there condition to early quit if "TRAVIS" env variable is not set. But it is still POC and yet it do not work, so it is just testing service for computing coverage for c++ project.

ancorgs: And is it working? Do we have some visible result anywhere?

jreidinger: @ancorgs not yet, I am still trying to make it working, but it is not high priority now.

jreidinger:
not enough time for more then year, so lets close it

PR #268, opened 2017-05-29
- work on DASDs

aschnell:
For https://trello.com/c/dnsteUCS/503-8-libstorageng-dasd-devices.


storage/Devices/DasdPtImpl.cc
```diff
@@ -0,0 +1,182 @@
...
+	return device_region.intersection(usable_region);
+    }
+
+
+    bool
+    DasdPt::Impl::equal(const Device::Impl& rhs_base) const
+    {
+	const Impl& rhs = dynamic_cast<const Impl&>(rhs_base);
+
+	if (!PartitionTable::Impl::equal(rhs))
```
shundhammer: Why not simply

    return PartitionTable::Impl::equal(rhs);

?

aschnell: Yes, or even remove the function. Depends on whether the class will have own data fields in the end.

shundhammer:
Besides that small thing, LGTM.

PR #270, opened 2017-05-31
- improve alignment for DASD partition tables

aschnell:
For https://trello.com/c/dnsteUCS/503-8-libstorageng-dasd-devices.

Might require minor changes in yast2-storage-ng (unit tests are fine).


storage/Utils/Alignment.cc
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+
+    Alignment::Alignment(const Alignment& alignment)
+	: impl(new Impl(alignment.get_impl()))
+    {
+    }
+
+
+    Alignment::~Alignment()
```
joseivanlopez: Is it necessary default destructor definition in this non virtual class?

aschnell: Kind of yes: The destructor of unique_ptr needs a complete type (see http://en.cppreference.com/w/cpp/memory/unique_ptr/%7Eunique_ptr), Alignment::Impl in this case. Since we do not provide AlignmentImpl.h in the API you would not be able to compile the class.

joseivanlopez: Ok, I see. Thanks.

PR #273, opened 2017-06-02
- implemented partition slots for DASD partition table

aschnell:
For https://trello.com/c/dnsteUCS/503-(8)-libstorageng%3A-dasd-devices.

### DIFF
--- a/storage/Utils/Algorithm.h
+++ b/storage/Utils/Algorithm.h
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2017 SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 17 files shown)


storage/Utils/Algorithm.h
```diff
@@ -0,0 +1,61 @@
...
+
+
+    /**
+     * Return the first missing number of the numbers in the vector starting
+     * from start. The numbers of the vector must be sorted and unique. The
+     * objects of the vector must have a get_number() function.
+     */
+    template<typename Type>
+    unsigned int
+    first_missing_number(const vector<Type>& values, unsigned int start)
```
joseivanlopez: To avoid errors, maybe the numbers can be sorted inside:
```
vector<unsigned int> numbers;
for(auto value : values) numbers.push_back(value->get_number());
   
for (auto n : sort(numbers.begin(), numbers.end()))
{
  if (number != n)
    return number;
  
  ++number;
}
```

aschnell: In general yes but currently it is not needed since both PartitionTable::get_partitions() and Md::get_all() sort by number.

BTW: sort() works in-place and does not return anything.

PR #280, opened 2017-06-11
Fix bug with compound actions formatter

joseivanlopez:
Avoid formatter error when only mounting.

PBI: https://trello.com/c/WzeqWCX8/518-5-storageng-simplified-summary.


aschnell:
LGTM

PR #281, opened 2017-06-12
- split directory examples into utils and examples

aschnell:
In the future the utils could even be packaged in a RPM.


joseivanlopez:
Utils also look like to be examples. Why not `examples/utils/` ?

aschnell:
To me they look more like generic utils now useful for e.g. debugging (e.g. display the probed and staging devicegraph found in a bug report) or probe a system and save the devicegraph.

Also the programs grow with option parsing and help texts which is not needed in simple examples.

PR #284, opened 2017-06-13
- work on probing

aschnell:
Partly for https://trello.com/c/4oghl4NL/558-libstorageng-multipath.

For multipath so far only the multipath devices themself are probed but not the partitions on them.


joseivanlopez:
To be honest, I do not know about multipath devices, but I guess that we can obtain several paths (and ids) from that devices. Is not there tests for that?

aschnell:
I have no tests for multipath yet since partitions on multipath are not probed yet. Thus any tests would be wrong.

In general multipath allows to combine several disks into one. The disks in this case are in reality all the same just accessed via several paths (e.g. several fiber channel wires).

joseivanlopez:
Ok. Thanks to explain me about multipath ;)

PR #289, opened 2017-06-27
- reduced interface of SystemCmd

aschnell:
The PR reduces the interface of SystemCmd to what is currently used. It also unifies the behaviour of the constructors (one even uses the other).

### DIFF
--- a/storage/Utils/SystemCmd.h
+++ b/storage/Utils/SystemCmd.h
@@ -30,7 +30,6 @@
 
 #include <string>
 #include <vector>
-#include <list>
 #include <boost/noncopyable.hpp>
 
 #include "storage/Utils/Exception.h"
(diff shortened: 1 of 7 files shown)


storage/Utils/SystemCmd.h
```diff
@@ -43,47 +42,79 @@ namespace storage
...
+
+	    /**
+	     * The command to be executed.
+	     */
+	    string command;
+
+	    /**
+	     * Should exceptions be thrown or not?
+	     */
+	    ThrowBehaviour throw_behaviour;
```
shundhammer: Please add a constructor that properly initializes those member variables that don't have a reasonable default constructor - like the ThrowBehaviour.

aschnell: The existing constructor does that. Do you additionally want a default constructor?

shundhammer: Argh. Forget it; I overlooked that one.

shundhammer:
LGTM

PR #291, opened 2017-07-03
- improved probing for multipath

aschnell:
For https://trello.com/c/4oghl4NL/558-3-libstorageng-multipath.

Although the patch is huge it is not complicated. The improved probing is explained in https://github.com/openSUSE/libstorage-ng/pull/291/files#diff-af4e787ec95e0be6fca8597c6d31f9e4. The rest follows from that. The larges part of the patch is updating the XML files in the unit tests.


testsuite/helpers/TsCmp.cc
```diff
@@ -41,6 +55,128 @@ namespace storage
...
+     * for identification, most others a uuid or a path. Some types are not
+     * handled at all, e.g. Nfs.
+     *
+     * The function makes assumptions that break in the general case and does
+     * no error checking. It can even ruin the devicegraph.
+     *
+     * Only enable it when you know what you are doing!
+     */
+    void
+    TsCmpDevicegraph::adjust_sids(const Devicegraph& lhs, Devicegraph& rhs) const
```
joseivanlopez: Why all this inaccessible code?

aschnell: Because sometimes it is needed: When the probing is modifies and the same objects are detected but in a different order. In that case the sids have to be adjusted. This is the case with the modifications I'm planning for multipath.

storage/Devices/BlkDeviceImpl.cc
```diff
@@ -208,9 +209,42 @@ namespace storage
...
+	for (Devicegraph::Impl::vertex_descriptor vertex : devicegraph->get_impl().vertices())
+	{
+	    const BlkDevice* blk_device = dynamic_cast<const BlkDevice*>(devicegraph->get_impl()[vertex]);
+	    if (blk_device && blk_device->get_impl().active)
+	    {
+		if (system_info.getCmdUdevadmInfo(blk_device->get_name()).get_majorminor() == majorminor)
+		    return true;
+	    }
+	}
+
```
aschnell: Yes, the same devices are interated. But the first iteration only does a match for the name in memory while the second does a real system lookup. The idea is to avoid the system lookup if the first iteration find a match.

joseivanlopez: Ok, I see. Thanks.

PR #298, opened 2017-07-07
- work on exceptions

aschnell:


ancorgs:
Is the swig catches list automagically generated from doxygen info? If so, it would be great to include the script or at least some instructions

aschnell:
I'm working on a script. But nothing so far that can be included or even used during make/build.

PR #299, opened 2017-07-07
- work on exceptions

aschnell:


shundhammer:
LGTM

For the record: We talked about listing those exceptions in the comment blocks in the headers; there is the risk of this getting outdated, but we both agreed that it's better to have that information there rather than in some SWIG description file far away from the relevant sources. We'll have to be careful to properly maintain those comment blocks when the exception behaviour changes.

PR #301, opened 2017-07-11
- added functions to set dasd type and format

aschnell:


shundhammer:
LGTM, but let's wait with the merge until that PPC build is done.

shundhammer:
LGTM again after the second commit.

PR #302, opened 2017-07-11
- work on exceptions

aschnell:


bindings/ruby/testsuite/function-call.rb
```diff
@@ -0,0 +1,24 @@
...
+  def test_function_call
+
+    environment = Storage::Environment.new(true, Storage::ProbeMode_NONE, Storage::TargetMode_DIRECT)
+    storage = Storage::Storage.new(environment)
+
+    assert_equal(storage.default_mount_by(), Storage::MountByType_UUID)
+    assert_raises(ArgumentError) { storage.default_mount_by("extra parameter") }
+    assert_raises(NoMethodError) { storage.does_not_exist() }
+
+    assert_equal(storage.public_send(:default_mount_by), Storage::MountByType_UUID)
```
joseivanlopez: Why is it necessary to test the same both: calling directly and using `public_send`? Is there any difference?

aschnell: I do not know if there is any difference. I just wanted to see that both work.

PR #303, opened 2017-07-13
Expose Alignment#grain

joseivanlopez:
Needed for PBI: https://trello.com/c/UBCLA6Qo/651-2-storageng-and-s-390-partitions-not-aligned.

The Proposal needs to know in advance the grain for alignment. With this info, the proposal is able to always generate aligned partitions, even for DASD. 

This PR should be merged before this one https://github.com/yast/yast-storage-ng/pull/312.

### DIFF
--- a/storage/Utils/Alignment.cc
+++ b/storage/Utils/Alignment.cc
@@ -85,6 +85,13 @@ namespace storage
     }
 
 
+    long
+    Alignment::offset() const
+    {
+	return get_impl().offset();
+    }
+
+
     unsigned long
     Alignment::grain() const
     {
--- a/storage/Utils/Alignment.cc
+++ b/storage/Utils/Alignment.cc
@@ -85,6 +85,13 @@ namespace storage
     }
 
 
+    unsigned long
+    Alignment::grain() const
+    {
+	return get_impl().calculate_grain();
+    }
+
+
     std::ostream&
     operator<<(std::ostream& s, const Alignment& alignment)
     {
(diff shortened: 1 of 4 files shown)


aschnell:
If you expose the grain please also expose the offset and always use both.

ancorgs:
@aschnell You are right. It makes sense to expose both. We currently don't use offset in the proposal algorithm for a reason. Can you please tell me if I'm right about it?

We don't use offset because all we do is to propose partitions within the given slots returned by `PartitionTable#unused_partitions_slots`, which starting point is already properly aligned. Our first proposed partition is always at the beginning of such slot and the sizes of all our partitions (with the only possible exception of the one place at the very end of the slot) are multiple of `slot.partition_table.alignment.grain`. The result is that our partitions are always aligned thanks to the correct initial alignment of the slot reported by libstorage.

Is that correct or we really need to include the offset in our calculations somehow?

aschnell:
In that case you should not need the offset.

Do you take care of the extra space needed for logical partitions?

ancorgs:
@aschnell Yes, we do it here https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/planned/assigned_space.rb#L136

Plan of the part for this PBI is stop using `disk.min_grain` everywhere and start using `partition_table.alignment.grain`. That includes the code linked above. Correct?

aschnell:
If ```disk.min_grain``` maps to ```topology.get_minimal_grain()``` the code was wrong since the minimal grain does not include any information about the real hardware topology. So using ```partition_table.alignment.grain``` is required in any case.

aschnell:
LGTM

PR #309, opened 2017-07-21
Bindings for some Region exceptions

joseivanlopez:


aschnell:
The file is generated. Please add comments to the functions and include the doxygen @throw tag.

The generation is not automated yet. For the time being just tell me to generate the file.

PR #316, opened 2017-08-01
- work on MD RAID

aschnell:
For https://trello.com/c/2tJ3LlpM/672-5-storageng-bios-raid-via-md-devices.


storage/Devices/MdImpl.cc
```diff
@@ -228,12 +230,27 @@ namespace storage
...
+	    else if (entry.has_container)
+	    {
+		MdMember* md_member = MdMember::create(prober.get_probed(), name);
+		md_member->get_impl().probe_pass_1a(prober);
+	    }
+	    else
+	    {
+		Md* md = Md::create(prober.get_probed(), name);
+		md->get_impl().probe_pass_1a(prober);
+	    }
```
joseivanlopez: Base class being aware about derived ones is a bit weird. Would it be possible different probe methods for each one?

storage/Devices/MdMemberImpl.cc
```diff
@@ -0,0 +1,166 @@
...
+	return tmp;
+    }
+
+
+    vector<const BlkDevice*>
+    MdMember::Impl::get_devices() const
+    {
+	vector<const BlkDevice*> tmp = Md::Impl::get_devices();
+
+	erase_if(tmp, [](const BlkDevice* blk_device) { return is_md_container(blk_device); });
```
joseivanlopez: Can the list of devices contains a parent?

storage/Devices/MdContainerImpl.cc
```diff
@@ -0,0 +1,122 @@
...
+
+    void
+    MdContainer::Impl::check() const
+    {
+	Md::Impl::check();
+
+	if (get_md_level() != MdLevel::CONTAINER)
+	    ST_THROW(Exception("invalid md-level for MdContainer"));
+
+	if (get_metadata() != "" && get_metadata() != "imsm" && get_metadata() != "ddf")
```
aschnell: Not after probing. But MdContainer::create() keeps it empty - neither imsm nor ddf is a good default. So e.g. in testsuites it may be empty.

storage/Devices/MdMemberImpl.cc
aschnell: The parent for a MdMember is the MdContainer and the BlkDevices the MdMember uses. MdMember::get_devices() returns only the BlkDevices, MdMember::get_md_container() returns only the MdContainer.

storage/Devices/MdImpl.cc
aschnell: Sure that would be possible but the logic would be there three times then.

PR #323, opened 2017-08-09
- simpler code

aschnell:
Simpler code, maybe not so obvious why it is simpler: get_md_container() looks at all parents, the new code avoids that.


shundhammer:
LGTM

PR #324, opened 2017-08-09
redirect package submissions to Factory

wfeldt:


package/libstorage-ng.changes
```diff
@@ -1,17 +1,4275 @@
--------------------------------------------------------------------
```
jreidinger: is such big change to changes required? it looks quite strange for me.

wfeldt: We didn't write any changes file so far.

jreidinger: I see there is some old entries in changes. If there is agreement on it, then I am fine with it.

wfeldt: Yes, that has to be sorted out in obs in the first submission.

shundhammer:
LGTM

PR #331, opened 2017-08-31
Configure snapper

shundhammer:
https://trello.com/c/ytK5Yoe4/554-5-storageng-installation-to-root-subvolume

There are no suitable commit callbacks where we could do this in Ruby code, so we discussed this and decided to move the first part into the C++ code.

This requires a flag to be set on the root Btrfs object that tells it to start configuring snapper at strategic points during commit: After the filesystem is created, after it is mounted, after it is added to /etc/fstab.

It is the Ruby code's responsibility to set that flag when appropriate.

This is done in https://github.com/yast/yast-storage-ng/pull/347

See also the old code where this was taken from:

https://github.com/yast/yast-storage/blob/master/src/modules/Storage.rb#L4849
https://github.com/shundhammer/yast-storage/blob/master/src/modules/StorageSnapper.rb#L74


Dockerfile
```diff
@@ -1,3 +1,4 @@
-FROM yastdevel/storage-ng
+FROM yastdevel/cpp
+RUN zypper --non-interactive install swig python-devel
```
lslezak: Add also the `libxml2-devel` package.

PR #335, opened 2017-09-06
- added function to query minimal number of devices for Md

aschnell:
For https://trello.com/c/YDRIZHHh/700-8-storageng-expert-partitioner-create-md-array.

### DIFF
--- a/storage/Devices/MdImpl.cc
+++ b/storage/Devices/MdImpl.cc
@@ -562,6 +562,32 @@ namespace storage
     }
 
 
+    unsigned int
+    Md::Impl::minimal_number_of_devices() const
+    {
+	switch (md_level)
+	{
+	    case MdLevel::RAID0:
+		return 2;
+
+	    case MdLevel::RAID1:
+		return 2;
+
+	    case MdLevel::RAID5:
+		return 3;
+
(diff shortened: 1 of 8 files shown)


storage/Devices/MdImpl.cc
```diff
@@ -562,6 +562,32 @@ namespace storage
...
+
+	    case MdLevel::RAID10:
+		return 2;
+
+	    default:
+		return 0;
+	}
+    }
+
+
```
joseivanlopez: For the RAID form in the expert partitioner, it would be necessary to known the number of devices for a RAID type before creating the RAID object, so maybe this method should be available as class method.

aschnell: Well, for that function this is easy. But for the function that calculates the size of the resulting RAID it is more complicated since that needs for more parameters, e.g. level, devices, parity, chunk size. And maybe those parameters even change in time, e.g. the size may depend on the metadata version.

What is the problem with creating the Md object?

joseivanlopez: Ok, I guess we can create the Md in the first step of the wizard and validate in that moment the number of devices.

PR #336, opened 2017-09-07
- changed some requires to suggests

aschnell:
For https://trello.com/c/hZHXGFuw/705-1-libstorageng-remove-dm-raid-dependency.


imobachgs:
The change LGTM. But can libstorage-ng crash if any of those tools is not installed?

aschnell:
Checking that is another task of the PBI.

PR #342, opened 2017-09-13
Fix changelog

shundhammer:
Fixed timestamp in commits so the change log checker doesn't complain:

https://ci.suse.de/job/libstorage-ng-master/31/console


shundhammer:
We decided to do this the right way; sntwint is fixing the changelog generator script.

Closing this PR for good.

PR #345, opened 2017-09-14
Update the scripts from openSUSE/linuxrc-devtools

lslezak:
- Update the scripts from openSUSE/linuxrc-devtools (already reviewed there!)
- **Generate the `.changes` file and check it in Travis**
- Install also `graphviz` at Travis (the build passed but there were some error messages)
- Use `.tar.xz` format for the source tarball  to be compatible with Jenkins and OBS packages
- Use a specific Travis build script, the tooling here is too different to the YaST packages,
it would requires several workarounds just for a single package. Using a specific Travis build script seems to be easier.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,26 @@
+#! /bin/sh
+
+# This is a CI build script for running inside the Travis builds.
+
+# exit on error immediately, print the executed commands
+set -e -x
+
+# generate the .tar.xz source tarball and the *.changes file
+utils/make_package
+
+# generate the *.spec file
+make -f Makefile.repo
+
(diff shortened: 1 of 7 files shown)


shundhammer:
LGTM

lslezak:
Merge postponed until the Docker images are rebuilt (https://hub.docker.com/r/yastdevel/cpp/builds/).

lslezak:
I have changed the Travis build to not use the YaST shared script, libstorage-ng needs some specific build preparations (generating the `.spec` and the `.changes`). I using a specific script seems to easier than adding workarounds there just for this single package.

PR #349, opened 2017-09-19
- provide functions to create/remove encryption while keeping children

aschnell:
For https://trello.com/c/ytK5Yoe4/554-5-storageng-installation-to-root-subvolume.

### DIFF
--- a/storage/Devices/BlkDeviceImpl.cc
+++ b/storage/Devices/BlkDeviceImpl.cc
@@ -478,17 +478,45 @@ namespace storage
     Encryption*
     BlkDevice::Impl::create_encryption(const string& dm_name)
     {
-	if (num_children() != 0)
-	    ST_THROW(WrongNumberOfChildren(num_children(), 0));
+	Devicegraph* devicegraph = get_devicegraph();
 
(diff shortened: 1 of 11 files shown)


storage/Devices/BlkDeviceImpl.cc
```diff
@@ -478,17 +478,45 @@ namespace storage
...
+	return encryption;
+    }
+
+
+    void
+    BlkDevice::Impl::remove_encryption()
+    {
+	Devicegraph* devicegraph = get_devicegraph();
+
+	Encryption* encryption = get_encryption();
```
shundhammer: I assume if there is no encryption layer on this level, this will throw an exception, right?
In that case: Do we have a corresponding has_encryption() predicate?

shundhammer: Disregard. I just found has_encryption() below.

aschnell: Yes, we have has_encryption() and get_encryption() will throw.

shundhammer:
LGTM

PR #351, opened 2017-09-22
- work on lvm thin provisioning

aschnell:
For https://trello.com/c/htNoDhHR/369-5-libstorage-ng-lvm-thin-provisioning-probing.


storage/Devices/LvmLv.h
```diff
@@ -62,6 +76,14 @@ namespace storage
 	 */
 	const LvmVg* get_lvm_vg() const;
 
+	/**
+	 * Create a logical volume with name lv_name and type lv_type in the
+	 * thin pool. Only supported lv_type is THIN.
+	 *
```
joseivanlopez: Would it be possible other types different to THIN? I mean, why is this parameter necessary?

aschnell: It parameter is likely needed when adding LVM snapshots or caches. So it is now here to avoid an API change and to have the options the same as for LvmVg::create_lvm_lv().

PR #365, opened 2017-10-05
Honor mkfs_options when creating filesystems

imobachgs:
Related to [bsc#1061289](https://bugzilla.suse.com/show_bug.cgi?id=1061289).


aschnell:
LGTM, although I do not see the relation to that bugzilla issue.

imobachgs:
Thanks! Actually, you are right. I've fixed the bug link.

PR #367, opened 2017-10-09
Commit options

imobachgs:


storage/Action.cc
```diff
@@ -61,7 +61,7 @@ namespace storage
 
 
 	void
-	Create::commit(CommitData& commit_data) const
+	Create::commit(CommitData& commit_data, CommitOptions& commit_options) const
```
aschnell: That parameter should be const.

storage/CommitOptions.h
```diff
@@ -0,0 +1,40 @@
...
+#define STORAGE_COMMIT_OPTIONS_H
+
+namespace storage
+{
+    class CommitOptions
+    {
+    public:
+
+	//CommitOptions(const bool force_rw);
+	CommitOptions(const bool force_rw): force_rw(force_rw) {};
```
aschnell: The const here is not really required since the parameter is passed by value.

storage/CommitOptions.h
```diff
@@ -0,0 +1,40 @@
...
+
+	~CommitOptions() {};
```
aschnell: The destructor should not be needed.

storage/Storage.h
```diff
@@ -264,6 +264,13 @@ namespace storage
...
 	 * @throw Exception
 	 */
+	void commit(CommitOptions& commit_options, const CommitCallbacks* commit_callbacks = nullptr);
+
+	/**
+	 * The actiongraph must be valid.
+	 *
+	 * @throw Exception
+	 */
 	void commit(const CommitCallbacks* commit_callbacks = nullptr);
```
aschnell: Should be marked as deprecated.

aschnell:
LGTM

imobachgs:
Thanks!

PR #371, opened 2017-10-19
- work on DASD support

aschnell:
For https://trello.com/c/A82hC5Sw/750-5-libstorageng-fba-dasd.


doc/dasd.md
```diff
@@ -0,0 +1,75 @@
...
+
+Can be used directly as a block device. But since the implicit partition stays
+this looks like a recipe for disaster.
+
+- Partition Types: Implicit, MS-DOS, GPT
+
+- Use as Block Device: Unsupported
+
+
+Glosar
```
lslezak: Typo: Glossary

doc/dasd.md
```diff
@@ -0,0 +1,75 @@
+
+DASDs
+-----
+
+There are ECKD and FBA DASDs.
```
lslezak: NP: I was wondering what all these abbreviations mean, later I noticed that there is a glossary at the bottom. Mabey you could add a note like `(See the glossary at the bottom.)`

doc/dasd.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Partition Types: None
+
+- Use as Block Device: No
+
+
+FBA DASDs
+---------
+
+Can be used with a MSDOS or GPT partition table. Otherwise implicit partition
```
lslezak: Question: I'm not a storage expert, but this looks a bit confusing to me. It can be used with MSDOS or GTP PT but cannot be used  as a block device? I thought an MSDOS partition *is* a block device... So how you can use it actually?

aschnell: The partitions on the MSDOS, GPT or imlicit partition table can be used as block devices. But the DASD itself cannot.

storage/Devices/PartitionableImpl.cc
```diff
@@ -118,7 +119,23 @@ namespace storage
 	    if (get_region().get_length() != parted.get_region().get_length())
 		ST_THROW(Exception("different size reported by kernel and parted"));
 
-	    PartitionTable* pt = create_partition_table(parted.get_label());
+	    PtType label = parted.get_label();
+
+	    // parted reports DASD partition type for implicit partition
```
joseivanlopez: NP: leftover partition word in "DASD partition type"?

aschnell: No, type should be table.

PR #386, opened 2017-11-07
- work on sorting of partitionables

aschnell:
For https://trello.com/c/6szyiBWI/763-2-sle15-storage-ng-select-the-disk-to-use-consistently.


aschnell:
Thanks for the review.

PR #387, opened 2017-11-14
- minor improvements

aschnell:


storage/Filesystems/MountPointImpl.cc
```diff
@@ -51,6 +51,10 @@ namespace storage
 	: Device::Impl(), path(path), mount_by(MountByType::DEVICE), freq(0), passno(0),
 	  active(true), in_etc_fstab(true)
     {
+#if 0
+	if (!valid_path(path))
+	    ST_THROW(InvalidMountPointPath(path));
+#endif
```
joseivanlopez: Why is this check commented? Similar below.

aschnell: The check fails in yast2-storage-ng. Looks as if first https://trello.com/c/NYy7AklQ/577-use-new-libstorage-ng-api-for-mount-points has to be done.

PR #394, opened 2017-11-23
Don't sort disks and dasds by default

ancorgs:
Part of https://trello.com/c/6szyiBWI/763-2-sle15-storage-ng-select-the-disk-to-use-consistently

This should not break the yast2-storage-ng testsuite IF this other PR is already merged there: https://github.com/yast/yast-storage-ng/pull/428


aschnell:
LGTM

PR #400, opened 2017-12-01
- work on device sorting

aschnell:


storage/Utils/AppUtil.h
```diff
@@ -106,6 +108,31 @@ void classic(StreamType& stream)
...
+     */
+    string
+    pad_front(const string& s, size_t w, char c);
+
+
+    struct NameSchema
+    {
+	NameSchema(regex re, size_t w, char c) : re(re), w(w), c(c) {}
+
+	const regex re;
```
ancorgs: Can I ask for more meaningful name for the fields of this struct. `re`, `w` and `c` are not the quintessence of clarity. :wink:

aschnell:
The disk and DASD /dev/vd thing is included in the unit tests.

Maybe I will add a test for the strict weak ordering in the future.

PR #406, opened 2017-12-14
- removed sorting of results

aschnell:


aschnell:
The travis problems are not caused by this pull request.

PR #411, opened 2017-12-18
Accept disk udev names starting by virtio- (bsc#1073254)

ancorgs:


ancorgs:
Looks like there is some reason to leave "virtio-" out. Let's wait for @aschnell to take a look.

ancorgs:
I did read https://github.com/openSUSE/libstorage-ng/blob/master/doc/udev.md but, to my limited understanding, it does not apply to "virtio-" (it's based on the device itself, not on its content). But, on the other hand, this comment from @aschnell https://bugzilla.suse.com/show_bug.cgi?id=1060569#c1 seems to imply that leaving "virtio-" out is intentional.

aschnell:
The email https://lists.freedesktop.org/archives/systemd-devel/2017-March/038397.html mentions that virtio names have changed multiple times so using them does not look as a very good idea.

Anyway, there is another generic problem: Since people can use any stupid udev link in fstab YaST
must be be able to find all. The library has extended functions (doing a real system lookup) for that
which can be exported in the API.

ancorgs:
I'm fine with whatever solution that allows YaST to identify those `virtio-SERIAL-part-NUMBER` devices.

That being said, as a user of the library API, I would expect `Device#udev_ids` to return all the known ids. If libstorage-ng wants to encourage the usage of stable ids (for example, only writing stable strings in fstab), I would have expected either an alternative function to get only those (let's say `Device#stable_udev_ids`) or an argument to the original function.

I would say that having the list filtered out on probing (which means it will not be, for example, in the xml dumps) and having to use a separate lookup function here and there is less intuitive than being able to trust the information contained in the device object. 

But again, I'm fine with whatever solution that allow us to move forward.

aschnell:
Different solution implemented in https://github.com/openSUSE/libstorage-ng/pull/414.

PR #414, opened 2017-12-19
- provide find_by_any_name in API

aschnell:
For bsc #1073254.


ancorgs:
This should be enough to fix bug#1073254. So LGTM from that point of view. But the packaging task seems to be broken according to travis.

aschnell:
I do not know why travis fails. These checks in travis have no value for me and I usually ignore failures. The package builds fine in OBS and IBS.

PR #419, opened 2017-12-20
Leave enough space even for 'reiserfs'

shundhammer:
When using filesystem type "reiserfs" the formatting looked ragged because it exceeded the previous max column width of 7. Now leaving some more room. Remember that this is only the maximum column width; if less width is required, it will use less width.


aschnell:
LGTM

PR #420, opened 2017-12-20
remove LIBVERSION_MAJOR macro from spec.in

wfeldt:
LIBVERSION_MAJOR will be 1 for the forseeable future.

Without this macro *.spec.in can be used directly as spec file for our
auto-submission process.


shundhammer:
LGTM

PR #428, opened 2018-01-18
- use /etc/fstab and /proc/mounts for probing NFS

aschnell:
For https://trello.com/c/R5gCe53x/215-3-libstorage-ng-use-fstab-crypttab-and-proc-or-sys-during-probing.


storage/Filesystems/NfsImpl.cc
```diff
@@ -100,30 +100,58 @@ namespace storage
...
 	// TODO the old library filters the mount options
 
-	vector<const FstabEntry*> nfs_entries = prober.get_system_info().getProcMounts().get_all_nfs();
-	for (const FstabEntry* fstab_entry : nfs_entries)
+	SystemInfo& system_info = prober.get_system_info();
+
+	/*
+	 * The key of the map is a pair of server and path of the NFS mounts.
+	 * The value of the map is a pair of vectors with the FstabEntries
+	 * from /etc/fstab and /proc/mounts.
```
mvidner: So the "gathering" part of this method allows multiple entries per key (and that's why it uses a std::vector instead of a std::optional), while the "processing" part assumes the vector will have at most 1 element as the PO does not want more?

aschnell: Yes. Additionally the API is now limited to one mount point per mountable.

PR #429, opened 2018-01-19
Add libstorage-ng-lang Recommends

zaitor:
Add missing libstorage-ng-lang Recommends

### DIFF
--- a/libstorage-ng.spec.in
+++ b/libstorage-ng.spec.in
@@ -53,6 +53,7 @@ BuildRequires:  swig >= 3.0.3
 BuildRequires:  pkgconfig(libxml-2.0)
 BuildRequires:  libjson-c-devel
 BuildRequires:  pkgconfig(python3)
+Recommends:     %{name}-lang
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 
 %description


aschnell:
LGTM. Thanks.

aschnell:
On the other hand there is already a "Requires: %{name}-lang" in line 74. Did you have any problems without the "Recommends"?

DimStar77:
Tjis change is wrong.

Since there is no libstorage-ng binary, the recommends in the preamble is useless

What he meant to do was replacing the requires in libstorage-ng1 with recommends (which is following what all pavkages do for lang)

aschnell:
I also just read the automatic reject mail. Reverting.

DimStar77:
> I also just read the automatic reject mail. Reverting.

Automatic? I'm not yet a bot :) but future might change that, who knows

DimStar77:
For reference, the intention was to fix

https://lists.opensuse.org/opensuse-factory/2018-01/msg00383.html

zaitor:
Just to confirm, that was the exact problem I was trying to fix - sorry about the cockup, that is what happens when you rush to do a quick edit in webview :-)

PR #431, opened 2018-01-20
Recommend, not Require lang package

zaitor:
Lang packages should only be recommended, not required - if they are mandatory, the split is pointless.


aschnell:
LGTM now.

PR #434, opened 2018-01-23
- find crypttab entries by correct block device

aschnell:
For https://trello.com/c/R5gCe53x/215-3-libstorage-ng-use-fstab-crypttab-and-proc-or-sys-during-probing.


storage/Devices/LuksImpl.cc
```diff
@@ -513,21 +514,15 @@ namespace storage
     void
     Luks::Impl::do_rename_in_etc_crypttab(CommitData& commit_data, const Device* lhs) const
```
joseivanlopez: Now `lhs` is not used at all.

aschnell: Yes, indeed.

Issue #440, opened 2018-01-31
ARM checks fail


ggardet:
armv7l and armv6l fails in OBS: https://build.opensuse.org/package/show/openSUSE:Factory:ARM/libstorage-ng
Because humanstring.test fails in 'make  check-TESTS'.

ggardet:
Here is the testsuite/Utils/humanstring.log:
`Running 4 test cases...
humanstring.cc(139): error: in "test_big_numbers": check test("en_GB.UTF-8", 16 * EiB - 1 * B, true, 2, true) == "16.00 EiB" has failed [16 EiB != 16.00 EiB] 
humanstring.cc(143): error: in "test_big_numbers": exception OverflowException expected but not raised
humanstring.cc(144): error: in "test_big_numbers": exception OverflowException expected but not raised

*** 3 failures are detected in the test module "libstorage"
FAIL humanstring.test (exit status: 201)
`

ggardet:
For the record: Bugzilla opened here: https://bugzilla.suse.com/show_bug.cgi?id=1078997

aschnell:
Fixed by https://github.com/openSUSE/libstorage-ng/pull/459.

PR #445, opened 2018-02-03
test: Run 'test_big_numbers' only if 'long double' is supported

ggardet:
It fixes ARM check: #440 
boo#1078997

### DIFF
--- a/testsuite/Utils/humanstring.cc
+++ b/testsuite/Utils/humanstring.cc
@@ -128,6 +128,7 @@ BOOST_AUTO_TEST_CASE(test_humanstring_to_byte)
 }
 
 
+#ifndef __NO_LONG_DOUBLE_MATH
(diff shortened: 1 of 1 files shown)


aschnell:
Good to see that it helps.

But please only disable those three tests that fail, not the complete function. Also add a check
for "16.01 EiB"  to overflow. That should still work.

ggardet:
As long as 'long double' is the same as 'double', all 'test_big_numbers' should be disabled since precision is not enough to handle them correctly. We have 15 digits of precision in decimal which is much more than 1 EiB. So, I think it does not make sense to enable those tests at all for ARM32.

ggardet:
@aschnell, what do you think about my last proposition? It is quite important, since it breaks ARM32 flavors.

aschnell:
Superseded by https://github.com/openSUSE/libstorage-ng/pull/459.

PR #450, opened 2018-02-06
Ensure not to write malformed /etc/fstab entries (bsc#1066763)

shundhammer:
This is the low-level fix to avoid writing /etc/fstab entries if vital things are missing. In this implementation, the offending entries are just not written to /etc/fstab, and an error is logged instead.


aschnell:
LGTM

PR #451, opened 2018-02-06
- work on error handling

aschnell:
See https://trello.com/c/AOfD3BfN/58-2-storageng-expert-partitioner-device-graph.


ancorgs:
Checked that with this `write_graphviz` propagates errors again. So YaST is again able to show a pop-up to the user to tell him something was wrong.

On the other hand, `save` is still quite resilient. Instead of raising the exception, it prints "I/O error : Permission denied" or "I/O error : No such file or directory" to stderr (depending on the case, of course) and throws no exception.

aschnell:
Apparently libxml needs special error handling. I will have to look at that some other day.

PR #453, opened 2018-02-08
- ignore disks without node in /dev

aschnell:
For https://bugzilla.suse.com/show_bug.cgi?id=1076971.


wfeldt:
If you include `<sys/stat.h>` anyway to get the `S_IS*` macros why don't you run stat() instead of an external program?

aschnell:
For remote probing and mockup in the testsuite.

wfeldt:
lgtm

PR #455, opened 2018-02-09
DO NOT MERGE test: fix changelog order automatically

jsrain:


wfeldt:
NO!

The git history is completely messed up in those travis builds! (And only in travis.) We must find out why. Fixing the log will just create a bunch of weird log entries.

jsrain:
You are right, I just wanted to give it a try (and am fine just closing/rejecting this PR).
The question for me is: Why does git2log behave so differently in the Travis environment? I don't think that Travis can take all the blame here - the environment is different, but git2log should be able to cope with it (like it does with different locale settings)

wfeldt:
I think it's the `--depth-50` travis uses when cloning the repo.

jsrain:
https://github.com/openSUSE/libstorage-ng/pull/456 seems to solve the problem.

PR #458, opened 2018-02-09
- cleanup

aschnell:


aschnell:
When I remove the rest of the deprecate functions the testsuite in yast2-storage-ng fails. We can look at that some other day.

PR #460, opened 2018-02-12
Removed unit tests that clash with new glibc number formatting

shundhammer:
https://bugzilla.suse.com/show_bug.cgi?id=1079855

Glibc now uses a narrow non-breakable space character (`U+202F`) as a thousands separator in certain locales. This makes some unit tests we did in the libstorage-ng  `humanstring()` function dependent on the version of glibc that it is linked against, and, worse, it's not even well-defined how to get that non-ASCII character properly to the compiler and through the unit test framework we use.

We decided to just drop those few unit tests for now. The important thing about testing that function is to check for proper use of Byte/kB/MB/GB etc. for most locales and Octet/kO/MO/GO etc. in French. We need to be able to trust glibc to format numbers correctly in general; all we add is the unit suffix.


aschnell:
LGTM

PR #462, opened 2018-02-13
- added function to immediately activate/deactivate mount point

aschnell:
For https://trello.com/c/s3Q03YDp/53-5-libstorage-ng-resizing-add-mount-handling.


storage/Filesystems/SwapImpl.cc
```diff
@@ -105,30 +107,16 @@ namespace storage
 
     void
     Swap::Impl::do_mount(CommitData& commit_data, const CommitOptions& commit_options,
```
joseivanlopez: Is this re-implementation needed? It looks like the same for `Mountable` base class.

storage/Filesystems/SwapImpl.cc
```diff
@@ -105,30 +107,16 @@ namespace storage
...
-	SystemCmd cmd(cmd_line);
-	if (cmd.retcode() != 0)
-	    ST_THROW(Exception("swapon failed"));
+	immediate_activate(mount_point);
     }
 
 
     void
-    Swap::Impl::do_umount(CommitData& commit_data, const MountPoint* mount_point) const
+    Swap::Impl::do_umount(CommitData& commit_data, MountPoint* mount_point) const
```
joseivanlopez: Same comment here.

storage/Filesystems/SwapImpl.cc
aschnell: Indeed, this is not needed anymore since the real code move to another function.

PR #475, opened 2018-02-22
- work on callbacks

aschnell:
For https://trello.com/c/suEMHOul/247-5-library-handle-errors-during-probing.

Needs changes to yast2-storage-ng (new callbacks parameter).

I recommend to ignore whitespaces during review.


shundhammer:
My gut feeling is that this is very much boilerplate code; there must be a way to avoid that much repetition. Mid-term, we should try to find a better way to do this (but I don't quite know why from the top of my head).

Meanwhile, LGTM.

aschnell:
Well, the code in Prober.cc needs many try-catch blocks. I though about writing a function that takes the two strings (progress message and potential error message) and another function that is executed. But that somehow looks also strange and at first less understandable.

On there other hand the boilerplate code to check the exit code of external programs vanished.

PR #479, opened 2018-02-27
- recalculate size of volume group when setting extent size

aschnell:


testsuite/lvm1.cc
```diff
@@ -157,3 +157,55 @@ BOOST_AUTO_TEST_CASE(chunk_size_too_small)
...
+
+    Disk* sda = Disk::create(staging, "/dev/sda", 1 * TiB);
+
+    LvmVg* lvm_vg = LvmVg::create(staging, "test");
+    lvm_vg->add_lvm_pv(sda);
+
+    LvmLv* normal1 = lvm_vg->create_lvm_lv("normal1", LvType::NORMAL, 1 * GiB);
+    LvmLv* normal2 = lvm_vg->create_lvm_lv("normal2", LvType::NORMAL, 6 * MiB);
+
+    BOOST_CHECK_EQUAL(lvm_vg->get_size(), 1 * TiB - 4 * MiB);
```
shundhammer: Some explanation or a `define` for those 4 MiB would be nice; I can only assume that this is some metadata size that LVM needs for its own management; but that could be clearer.

testsuite/lvm1.cc
```diff
@@ -157,3 +157,55 @@ BOOST_AUTO_TEST_CASE(chunk_size_too_small)
...
+    BOOST_CHECK_EQUAL(normal1->get_size(), 1 * GiB);
+    BOOST_CHECK_EQUAL(normal2->get_size(), 4 * MiB);
+
+    lvm_vg->set_extent_size(128 * MiB);
+
+    BOOST_CHECK_EQUAL(lvm_vg->get_size(), 1 * TiB - 128 * MiB);
```
shundhammer: Similar here.

shundhammer:
One small documentation issue, otherwise LGTM.

PR #481, opened 2018-02-28
- always allow to activate multipath (see bsc#1082542)

aschnell:
Requires small adjustments in yast2-storage-ng and yast2-autoinstallation.


aschnell:
I already have patches. I can send them to you. Maybe I missed something.

PR #483, opened 2018-03-01
- minor fixes and improvements

aschnell:


storage/Devices/MultipathImpl.cc
```diff
@@ -110,14 +110,16 @@ namespace storage
 	{
 	    bool looks_like_real_multipath = CmdMultipath(true).looks_like_real_multipath();
 
+	    y2mil("looks-like-real-multipath:" << looks_like_real_multipath);
+
 	    if (!activate_callbacks->multipath(looks_like_real_multipath))
 	    {
-		y2mil("user canceled activation of multipath");
+		y2mil("user does want multipath");
```
joseivanlopez: Is not the text inverted? I mean, I guess here it should be "user does not want multipath".

PR #487, opened 2018-03-07
- handle duplicate LVM PVs during probe

aschnell:
This will be needed to avoid errors when trying to modify LVM VGs that are broken (have missing or duplicate PVs).


shundhammer:
LGTM

PR #488, opened 2018-03-08
- minor improvements

aschnell:


imobachgs:
Additionally, you could add bug number #1079000 to avoid submission problems with the build service.

PR #489, opened 2018-03-12
- allow to remove partial volume group

aschnell:


storage/EtcCrypttab.cc
```diff
@@ -239,7 +239,7 @@ namespace storage
 		if (boost::starts_with(blk_device, "LABEL=") && blk_device.substr(6) == label)
 		    return entry;
 
-		if (boost::starts_with(blk_device, "/dev/disk/by-label/") && blk_device.substr(19) == label)
+		if (boost::starts_with(blk_device, "/dev/disk/by-label/") && blk_device.substr(19) == udev_encode(label))
```
shundhammer: That numeric constant 19 in the code looks like an accident waiting to happen.

How about a generic string function that checks for a prefix ("/dev/disk/by-label") and returns the remainder of the string after that prefix?

shundhammer: Let's leave it like this for now, but let's think about how such a function could work.

shundhammer:
Other than that one remark, LGTM.

shundhammer:
Looks even better to me. :+1:

PR #490, opened 2018-03-12
- handle erroneous small disks in get_unused_partition_slots()

aschnell:


aschnell:
The patch causes failures in yast2-storage-ng. Needs more research.

aschnell:
The patch was fine. The failure was caused by other outdated code on my system.

Created new PR https://github.com/openSUSE/libstorage-ng/pull/492.

PR #492, opened 2018-03-13
 - handle erroneous small disks in get_unused_partition_slots()

aschnell:


storage/Utils/RegionImpl.cc
```diff
@@ -71,7 +71,10 @@ namespace storage
     unsigned long long
     Region::Impl::get_end() const
     {
-	// TODO check if region is empty
+#if 0
+	if (empty())
+	    ST_THROW(Exception("empty region"));
+#endif
```
joseivanlopez: Why is it not possible to enable this check?

aschnell: Because it breaks tests in yast2-storage-ng.

PR #496, opened 2018-03-16
- do not recognize 0x1f as extended partition

aschnell:


rw4s:
LGTM.  This should definitely match the kernel's PoV.

PR #497, opened 2018-03-19
Enable Region#end exception

joseivanlopez:
Part of PBI: https://trello.com/c/jPe9PnBi/308-3-sles15-p1-1083887-zero-size-disks-not-able-to-find-the-scsi-disks-while-trying-to-install-sles-15-on-vmax-luns


aschnell:
LGTM

PR #502, opened 2018-03-26
Simplified supplements for lang subpkg (bsc#1081454)

shundhammer:
https://bugzilla.suse.com/show_bug.cgi?id=1081454

Keep it simple: Use a very plain "supplements" to get the lang subpackage installed by default (the user can still choose to not install it without getting a dependency conflict).


wfeldt:
lgtm

PR #503, opened 2018-03-27
- only insert resize action for devices existing in LHS and RHS

aschnell:
For https://trello.com/c/6DxcCIgy/334-not-planned-sles15-p1-1085523-resizing-and-formatting-a-partition-crashes-autoyast.


testsuite/dependencies/resize/resize4.cc
```diff
@@ -0,0 +1,22 @@
...
+
+
+using namespace storage;
+
+
+BOOST_AUTO_TEST_CASE(dependencies)
+{
+    set_logger(get_stdout_logger());
+
+    //
```
joseivanlopez: NP: it would be good an explanation as in other resize tests. Otherwise, I would simply remove the comment.

testsuite/dependencies/resize/resize5.cc
```diff
@@ -0,0 +1,22 @@
...
+
+
+using namespace storage;
+
+
+BOOST_AUTO_TEST_CASE(dependencies)
+{
+    set_logger(get_stdout_logger());
+
+    //
```
joseivanlopez: NP: The same here.

PR #507, opened 2018-04-05
- added function to set freq and passno

aschnell:
For https://trello.com/c/yzueuWcq/256-1-propose-correct-better-mount-options-for.

### DIFF
--- a/storage/Devices/Encryption.cc
+++ b/storage/Devices/Encryption.cc
@@ -63,7 +63,7 @@ namespace storage
 
 
     void
-    Encryption::set_mount_by(const MountByType mount_by)
+    Encryption::set_mount_by(MountByType mount_by)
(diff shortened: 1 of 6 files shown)


shundhammer:
I would really like to have alternative names for "freq" and "passno" - which one is the fsck pass, which one is the pass for that ancient "dump" backup tool that fell out of use after the late 70s?

Even after having written two separate parsers for /etc/fstab I'd have to look this up in "man fstab". I know that the first number is that (nowadays completely useless) dump pass and the last one is the one that still has a use to day: The fsck pass.

Please add at least aliases for them, for both getters and setters. If anybody wants those weird names because it's called like that in "man fstab", okay, but those names really always only confuse people using them.

Other than that, LGTM.

PR #508, opened 2018-04-05
- support lock within same process

aschnell:


storage/Utils/Lock.h
```diff
@@ -45,7 +45,10 @@ namespace storage
 	LockException(pid_t locker_pid);
 
 	/**
-	 * pid of process holding lock or 0 if it could not be determined.
+	 * pid of one of the process holding a lock. The oid is 0 if it could
```
joseivanlopez: Typo: oid

PR #510, opened 2018-04-10
Expose the vfstype field from fstab (bsc#1088426)

ancorgs:
Needed by https://github.com/yast/yast-storage-ng/pull/599 to detect and correct entries which use `nfs4` in vfstype.


shundhammer:
LGTM if Travis etc. are happy.

PR #511, opened 2018-04-12
Add MountPoint#mount_type to XML format (follow up on bsc#1088426)

ancorgs:
Related to #510 (in the scope of https://trello.com/c/kXnNLZPd/244-5-nfsv4x-support). That pull request added the field in the class. This makes it available in the XML format.


joseivanlopez:
Should not it be also added to `MountPointImpl#save` ?

PR #516, opened 2018-04-17
- fix unused partition slots calculation for logical partitions (bsc#1088483)

aschnell:


shundhammer:
LGTM

ancorgs:
This only breaks one test in yast-storage-ng, I will take a look. It should be easy to fix.

But it doesn't pass the regression test we prepared for the original bug report. See this comment https://trello.com/c/A2IzetTP/349-2-sles15-p1-1088483-storage-ng-parted-error-about-wrong-partition-end-sector#comment-5ad68e23ecf4d067cdaa1e92

PR #519, opened 2018-04-20
update git2log script to avoid identical timestamps in changelog

wfeldt:
This is a copy of the updated `git2log` script from https://github.com/openSUSE/linuxrc-devtools/pull/16


aschnell:
LGTM

PR #522, opened 2018-04-26
- centralized reading /sys/block

aschnell:
For https://trello.com/c/XH6Ig5kp/376-3-leap15-impossible-to-install-on-iscsi-or-fcoe.


storage/Utils/LightProbe.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace storage
+{
+
+    /**
+     * This function does some very light probing and returns true iff some
+     * disks or DASDs were found in the system. No devicegraph is generated.
+     * The exact type of devices probed may change.
+     *
+     * @throw Exception
```
joseivanlopez: Is it needed to return an exception? I guess it simply could return `false` is something goes wrong?

aschnell: Yes, it could. But maybe the caller wants to distinguish those cases.

PR #527, opened 2018-05-15
Use stricter NTFS max label size (bsc#1084867)

shundhammer:
This is for https://trello.com/c/pVvQMTRq/313-2-sles15-p3-1084867-installer-does-not-handle-xfs-label-length-limitations

to fix https://bugzilla.suse.com/show_bug.cgi?id=1084867 and possibly others that might be upcoming

`man ntfslabel` says max label size is 128 for NTFS, but all other sources I found say it's just 32:

- https://www.lifewire.com/volume-label-2626045 (found by locilka)
- https://en.wikipedia.org/wiki/Volume_(computing)#Volume_label
- https://stackoverflow.com/questions/3704109/what-is-the-maximum-length-of-a-windows-xp-drive-label

### DIFF
--- a/storage/Filesystems/NtfsImpl.h
+++ b/storage/Filesystems/NtfsImpl.h
@@ -54,7 +54,7 @@ namespace storage
 	virtual bool supports_unmounted_grow() const override { return true; }
 
 	virtual bool supports_label() const override { return true; }
-	virtual unsigned int max_labelsize() const override { return 128; }
+	virtual unsigned int max_labelsize() const override { return 32; }
 
 	virtual bool supports_uuid() const override { return true; }
 


aschnell:
LGTM

storage/Filesystems/NtfsImpl.h
```diff
@@ -54,7 +54,7 @@ namespace storage
 	virtual bool supports_unmounted_grow() const override { return true; }
 
 	virtual bool supports_label() const override { return true; }
-	virtual unsigned int max_labelsize() const override { return 128; }
+	virtual unsigned int max_labelsize() const override { return 32; }
```
ancorgs: It would have been nice to have the rationale with all those pointers somewhere close to this 32. Or at least a pointer to the bug report and add the information there.

So far, the useful information and the links to all the info sources are only in the pull request description, which is not the most trustworthy information in the long term (neither the easier to find while debugging).

shundhammer: That Stack Exchange post contained a link to a Microsoft site, but that link has become invalid in the meantime.

I also would have preferred to have a more authoritative source, preferably from Microsoft themselves, but I couldn't find anything about that NTFS label length on their web pages; but there are literally dozens of third-party sites all claiming it's just 32.

So I chose to remain on the conservative side and rather use the shorter limit. It's not that this is a painful restriction; most other filesystem types have a much shorter max size (typically 12-16). The only exception is Btrfs.

The most prominent use of those labels from our perspective is in /etc/fstab anyway, and more than 32 seems pretty excessive to me already.

PR #533, opened 2018-06-04
- merged po files

aschnell:


po/af.po
```diff
@@ -1601,6 +1601,11 @@ msgstr "Skep partisie %1$s (%2$s)"
 msgid "Probing partitions failed"
 msgstr ""
 
+#. TRANSLATORS: error message
+#, fuzzy, c-format
+msgid "Probing partitions on %s failed"
+msgstr "Skep partisie %1$s (%2$s)"
```
jreidinger: this looks wrong and broken

aschnell: The string is marked as fuzzy and AFAIK not included in the mo file.

PR #535, opened 2018-06-07
Use proper exception

joseivanlopez:
Needed for PBI: https://trello.com/c/OGssTRzO/393-osdistribution-p2p1-1094963-opensuse-150-installation-or-upgrade-fails-if-disk-contains-encrypted-partition-luks

Now, a `DeviceNotFoundByName` is raised when `BlkDevice#find_by_any_name` fails. General `Exception` could still be raised when a wrong devicegraph is used.

### DIFF
--- a/Makefile.ci
+++ b/Makefile.ci
@@ -6,7 +6,7 @@
 .PHONY: package
 package: obs-package-from-git
 	./$^ \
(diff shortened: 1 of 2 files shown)


aschnell:
LGTM

PR #542, opened 2018-07-09
- added support for f2fs (fate#325755)

aschnell:
For https://hackweek.suse.com/17/projects/add-features-to-libstorage-ng.


aschnell:
Thanks for the review.

BTW: supports_modify_label() is not added to the API (at least no the API useable for others) since it is only present in *Impl.{h,cc} files.

PR #549, opened 2018-07-26
- added special handling when creating udev ids starting with dm-uuid for partitions on multipath (bsc#1099394)

aschnell:
For https://trello.com/c/Ncr2zQi3/126-sles12-sp3-p2-l3-1099394-yast-constructs-wrong-udev-name-for-dm-uuid-names.

### DIFF
--- a/storage/Devices/PartitionImpl.cc
+++ b/storage/Devices/PartitionImpl.cc
@@ -21,6 +21,8 @@
  */
 
 
+#include <boost/algorithm/string.hpp>
+
 #include "storage/Utils/AppUtil.h"
 #include "storage/Utils/SystemCmd.h"
 #include "storage/Utils/StorageDefines.h"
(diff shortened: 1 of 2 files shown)


storage/Devices/PartitionImpl.cc
```diff
@@ -457,16 +459,28 @@ namespace storage
...
 	for (const string& udev_id : partitionable->get_udev_ids())
-	    udev_ids.push_back(udev_id + postfix);
+	{
+	    // The partition link for udev id links starting with dm-uuid is
+	    // special for multipath and dmraid (see bsc #1099394). Handling
+	    // dmraid here is optional since the links are not whitelisted.
+
+	    if (boost::starts_with(udev_id, "dm-uuid-mpath") ||
+		boost::starts_with(udev_id, "dm-uuid-DMRAID"))
+		udev_ids.push_back("dm-uuid" + addition + udev_id.substr(7));
```
lslezak: Where the magic value `7` comes from? Something like `sizeof("dm-uuid")` would be much better readable...

aschnell: But could be evaluated at runtime instead of compile time.

aschnell: ```sizeof("dm-uuid")``` is wrong since it included the trailing zero byte.

jreidinger: You should believe more in your compiler that it optimize this trivial constant string and actually. It is really done, see https://gcc.godbolt.org/#%7B%22version%22%3A3%2C%22filterAsm%22%3A%7B%22labels%22%3Atrue%2C%22directives%22%3Atrue%2C%22commentOnly%22%3Atrue%2C%22intel%22%3Atrue%2C%22colouriseAsm%22%3Atrue%7D%2C%22compilers%22%3A%5B%7B%22sourcez%22%3A%22MQSwdgxgNgrgJgUwAQB4IGcAuAncBzAPgCgsBDTECJCAezC2oAtTsAqJAawDEaakBeJACIAQiyEBuIkXCYkAZVIAPAMKkoUMRA4AKCMzZJ0ygPpZcYPAEoiAbyJIkIAGZIdRnJAC2ABx3GlMxx8ABpOHhow8ygEMB1uXiskJP5BAAYbR3tHR2wETBhsMCQ0qUcAX2lc%2FMLigEYpcqAA%3D%22%2C%22compiler%22%3A%22g453%22%2C%22options%22%3A%22-O2%22%7D%5D%7D

so yes, gcc with -O2 optimize it out and adds number to assembler.

PR #551, opened 2018-08-01
- added class to represent Xen Virtual Partitions

aschnell:
For https://trello.com/c/EqN6jm0m/172-5-sles15-p3-1085134-cannot-install-when-using-xen-virtual-partitions-xvda1xvda2.


storage/Utils/AppUtil.h
```diff
@@ -132,22 +134,18 @@ void classic(StreamType& stream)
      */
     struct NameSchema
```
mvidner: This class does not have a description and neither does format_to_name_schemata
What is this good for? Maybe a couple of examples will be enough?

Extracting a numeric/alphabetic id (or several ids) out of a device name? OK, but what is the padding good for?

storage/Devices/StrayBlkDeviceImpl.cc
```diff
@@ -0,0 +1,138 @@
...
+    StrayBlkDevice::Impl::Impl(const xmlNode* node)
+	: BlkDevice::Impl(node)
+    {
+    }
+
+
+    string
+    StrayBlkDevice::Impl::get_pretty_classname() const
+    {
+	// TRANSLATORS: name of object
```
mvidner: Translators will have a hard time translating our internal abstraction to something meaningful, They will not even know that it is our made-up term. Please give them some hints, maybe copy the class description.

storage/Devices/StrayBlkDevice.h
```diff
@@ -0,0 +1,116 @@
...
+     * This class represents stray block devices. Stray block devices
+     * have no parent in the devicegraph and cannot be partitioned.
+     *
+     * Normally only Disks and Dasds have no parents in the
+     * devicegraph but Disks and Dasds can be partitioned.
+     *
+     * Currently only Xen Virtual Partitions are represented as stray
+     * block devices (e.g. /dev/xvda1). Note that partitionable
+     * devices in Xen are represented as Disks (e.g. /dev/xvda) and
+     * partitions on them as normal Partitions (e.g. also /dev/xvda1).
```
mvidner: So, while we are adding this for the case when xvda1 has no parent, OTOH in some deployments, xvda1 is a partition on a xvda disk, just like with sda and vda. Right?

storage/Utils/AppUtil.h
aschnell: Both have a description. Anyway I can add an example for the usage.

storage/Devices/StrayBlkDeviceImpl.cc
aschnell: Seems difficult without repeating the explanation over and over since the term is already used four times.

storage/Utils/AppUtil.h
```diff
@@ -132,28 +134,30 @@ void classic(StreamType& stream)
...
 
     /**
-     * Formats the string s to the first matching name schema: All sub-matchs
+     * Formats the string s to the first matching name schema: All sub-matches
      * will be padded as defined in the name schema.
+     *
+     * This can be used to sort device names. E.g. if it is desired to
+     * have "sdb" before "sdaa" the names can be transformed to "sd b"
+     * and "sdaa" and be sorted alphabetical. Used by
+     * Device::get_sort_key().
```
mvidner: Thanks for the example!

PR #552, opened 2018-08-02
Do not crash when displaying summary for an encrypted but not mounted disk (bsc#1099181)

lslezak:
## Fix

- Do not call `filesystem->get_mount_point()->get_path()`, it throws an exception when the mount point is not defined. Moreover the parameter is not used in the formatted message.
- Added unit test

### Additional Improvements

When I wanted to compile libstorage-ng locally I got a compile error about a missing JSON header file. It was not obvious what was missing in my system. It turned out that `libjson-c-devel` package is needed for compiling.

- Added `libjson-c-devel` package to the main README.md
- Added an autoconf check to fail early, it also prints a suggestion which package to install
- Use the compiler and linker flags from the `json-c` pkgconfig
- Updated the `build_dependencies:*` Rake tasks to read the alternative spec file when `.obsdir` does not exist yet

## The Original Crash

storage_summary_crash

## With the Fix Applied

storage_summary_fixed

### DIFF
--- a/testsuite/CompoundAction/encrypted-sentence.cc
+++ b/testsuite/CompoundAction/encrypted-sentence.cc
@@ -1,0 +1,35 @@
+
+#define BOOST_TEST_DYN_LINK
+#define BOOST_TEST_MODULE libstorage
+
+#include <boost/test/unit_test.hpp>
+
+#include "storage/Filesystems/Ext4.h"
+#include "storage/Devices/Encryption.h"
+
+#include "testsuite/CompoundAction/Fixture.h"
+
+using namespace storage;
+
+BOOST_FIXTURE_TEST_SUITE(btrfs_subvolume_sentence, test::CompoundActionFixture)
+
+// test for https://bugzilla.suse.com/show_bug.cgi?id=1099181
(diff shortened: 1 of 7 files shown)


aschnell:
I do not think it is a good idea to first add the unit test and then the fix. This will make it hard to bisect the project.

lslezak:
The usual TDD approach is to implement a failing test and then fix the code to make it pass. This ensures that the test and the fix are correct. But I see your point so I have swapped the commits.

testsuite/CompoundAction/encrypted-sentence.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "storage/Devices/Encryption.h"
+
+#include "testsuite/CompoundAction/Fixture.h"
+
+using namespace storage;
+
+BOOST_FIXTURE_TEST_SUITE(btrfs_subvolume_sentence, test::CompoundActionFixture)
+
+// test for https://bugzilla.suse.com/show_bug.cgi?id=1099181
+BOOST_AUTO_TEST_CASE(test_sentence_on_entrypted_not_mounted)
```
joseivanlopez: NP: typo "entrypted"

PR #554, opened 2018-08-03
Inactive raid

mvidner:
- https://trello.com/c/8kPdHfZR
- https://bugzilla.suse.com/show_bug.cgi?id=1090010

#### Situation

2 of 3 devices in a RAID5 array have gone missing. So, the RAID is "inactive", and it is only available as /dev/md*NUMBER*, not as /dev/md/*NAME*.

#### Before this PR:
raid-inactive-before1
Details:
 raid-inactive-before2

#### With this PR
Here's an inactive RAID (made of /dev/sde1 and some other, missing, devices)
raid-inactive

For comparison, here is a regular active RAID (made of /dev/sd[abc]1)
raid-active

The "Active: Yes/No" field under the RAID heading is done in https://github.com/yast/yast-storage-ng/pull/717

### DIFF
--- a/storage/Devices/MdImpl.cc
+++ b/storage/Devices/MdImpl.cc
@@ -375,6 +375,7 @@ namespace storage
     Md::Impl::probe_mds(Prober& prober)
     {
 	SystemInfo& system_info = prober.get_system_info();
+	const MdLinks& md_links = system_info.getMdLinks();
 
 	for (const string& short_name : prober.get_sys_block_entries().mds)
 	{
@@ -383,8 +384,20 @@ namespace storage
 	    try
 	    {
 		const MdadmDetail& mdadm_detail = system_info.getMdadmDetail(name);
+
 		if (!mdadm_detail.devname.empty())
-		    name = DEV_MD_DIR "/" + mdadm_detail.devname;
+		{
+		    MdLinks::const_iterator it = md_links.find(short_name);
+		    if (it != md_links.end())
+		    {
+			// the mapping is backwards so we must iterate the result
+			const vector<string>& links = it->second;
+			if (std::find(links.begin(), links.end(), mdadm_detail.devname) != links.end())
+			{
+			    name = DEV_MD_DIR "/" + mdadm_detail.devname;
+			}
+		    }
+		}
 
 		const ProcMdstat::Entry& entry = system_info.getProcMdstat().get_entry(short_name);
 
--- a/storage/Devices/MdImpl.cc
+++ b/storage/Devices/MdImpl.cc
@@ -414,6 +414,7 @@ namespace storage
 		else
 		{
 		    Md* md = Md::create(prober.get_system(), name);
+		    md->get_impl().set_active(!entry.inactive);
 		    md->get_impl().probe_pass_1a(prober);
 		}
 	    }
--- a/storage/SystemInfo/DevAndSys.cc
+++ b/storage/SystemInfo/DevAndSys.cc
@@ -221,7 +221,16 @@ namespace storage
 
     MdLinks::MdLinks()
     {
-	map<string, string> links = getDirLinks("/dev/md");
+        map<string, string> links;
+	try
+	{
+	    links = getDirLinks("/dev/md");
+	}
+	catch (const Exception&)
+	{
+	    // OK, no /dev/md at all
+	}
+
 	for (const map<string, string>::value_type& it : links)
 	{
 	    string::size_type pos = it.second.find_first_not_of("./");
--- a/testsuite/SystemInfo/mdlinks.cc
+++ b/testsuite/SystemInfo/mdlinks.cc
@@ -2,8 +2,8 @@
 #define BOOST_TEST_DYN_LINK
 #define BOOST_TEST_MODULE libstorage
 
+#include <numeric>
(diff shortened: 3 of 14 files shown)


storage/Devices/MdImpl.cc
```diff
@@ -371,6 +372,15 @@ namespace storage
     }
 
 
+    static
+    bool file_exists(const string& pathname)
```
mvidner: we should probably use `boost::filesystem::exists`
see also `checkDir` and `checkNormalFile` in AppUtil.cc

aschnell: I would wait until we can use C++17 where filesystem is included.

storage/Devices/MdImpl.cc
```diff
@@ -383,8 +393,13 @@ namespace storage
 	    try
 	    {
 		const MdadmDetail& mdadm_detail = system_info.getMdadmDetail(name);
+
 		if (!mdadm_detail.devname.empty())
-		    name = DEV_MD_DIR "/" + mdadm_detail.devname;
+                {
+		    string devname = DEV_MD_DIR "/" + mdadm_detail.devname;
+                    if (file_exists(devname))
```
aschnell: During probing everything has to use SystemInfo for mocking in the testsuite and remote probing.

testsuite/SystemInfo/mdlinks.cc
```diff
@@ -33,6 +33,27 @@ check(const vector<string>& input, const vector<string>& output)
...
+
+    ostringstream parsed;
+    parsed.setf(std::ios::boolalpha);
+    parsed << mdlinks;
+
+    string lhs = parsed.str();
+    vector<string> output_n;
+    std::transform(output.begin(), output.end(), std::back_inserter(output_n),
+                   [](const string & s) { return s + '\n'; });
+    string rhs = boost::join(output_n, "");
```
aschnell: It is possible to use

```boost::join(v, "-") + (v.empty() ? ""s : "-"s)```

or

```std::accumulate(v.begin(), v.end(), ""s, [](const string& a, const string& b) { return a + b + "-"; });```

File: storage/SystemInfo/DevAndSys.cc [hunk 4]
```diff
@@ -221,7 +221,16 @@ namespace storage
 
     MdLinks::MdLinks()
     {
-	map<string, string> links = getDirLinks("/dev/md");
+        map<string, string> links;
+        try
+        {
+            links = getDirLinks("/dev/md");
+        }
+        catch (Exception& e)
```
aschnell: It is common to catch exception by const reference.

storage/Devices/MdImpl.cc
```diff
@@ -383,8 +384,20 @@ namespace storage
 	    try
 	    {
 		const MdadmDetail& mdadm_detail = system_info.getMdadmDetail(name);
+
 		if (!mdadm_detail.devname.empty())
-		    name = DEV_MD_DIR "/" + mdadm_detail.devname;
+                {
+                    auto it = md_links.find(short_name);
```
ancorgs: In the rest of the libstorage-ng code, when an iterator is declared the explicit class is always used instead of `auto`[1]. For consistency, I would use the whole class name here.

[1] Easily checked by running

```
grep -r " it =" storage/
```

storage/Devices/MdImpl.cc
```diff
@@ -401,6 +414,7 @@ namespace storage
 		else
 		{
 		    Md* md = Md::create(prober.get_system(), name);
+                    md->get_impl().set_active(!entry.inactive);
```
ancorgs: looks wrongly indented here in Github.

storage/Devices/MdImpl.cc
```diff
@@ -375,6 +375,7 @@ namespace storage
     Md::Impl::probe_mds(Prober& prober)
     {
 	SystemInfo& system_info = prober.get_system_info();
+        const MdLinks& md_links = system_info.getMdLinks();
```
ancorgs: Looks wrongly indented here at Github.

storage/SystemInfo/DevAndSys.cc
```diff
@@ -221,7 +221,16 @@ namespace storage
 
     MdLinks::MdLinks()
     {
-	map<string, string> links = getDirLinks("/dev/md");
```
ancorgs: More indentation problems?

storage/SystemInfo/DevAndSys.cc
```diff
@@ -221,7 +221,16 @@ namespace storage
...
+        map<string, string> links;
+	try
+	{
+	    links = getDirLinks("/dev/md");
```
aschnell: Use DEV_MD_DIR.

storage/Devices/MdImpl.cc
```diff
@@ -401,6 +414,7 @@ namespace storage
 		else
 		{
 		    Md* md = Md::create(prober.get_system(), name);
+		    md->get_impl().set_active(!entry.inactive);
```
aschnell: Is that correct? The ```active``` flag of BldDevice::Impl "means that the device node exists". That is the case even for RAIDs where the MD subsystem reports them RAID as inactive.

PR #555, opened 2018-08-06
- fix build with glibc 2.28

aschnell:


mvidner:
I would have quoted this part of the 2.28 release notes in the commit message, but this is fine.

> * The macros 'major', 'minor', and 'makedev' are now only available from
  the header <sys/sysmacros.h>; not from <sys/types.h> or various other
  headers that happen to include <sys/types.h>.  These macros are rarely
  used, not part of POSIX nor XSI, and their names frequently collide with
  user code; see https://sourceware.org/bugzilla/show_bug.cgi?id=19239 for
  further explanation.
>
>    <sys/sysmacros.h> is a GNU extension.  Portable programs that require
  these macros should first include <sys/types.h>, and then include
  <sys/sysmacros.h> if __GNU_LIBRARY__ is defined.

PR #559, opened 2018-08-08
- add support for exFAT

aschnell:
Not for any PBI that I know of.

I noticed that openSUSE supports exFAT while reading https://old.reddit.com/r/openSUSE/comments/94ci2d/exfat_support/. In the past it was not supported (see https://features.opensuse.org/316872).
So we can add support in YaST for exFAT now as well.


storage/Filesystems/ExfatImpl.cc
```diff
@@ -0,0 +1,100 @@
...
+    Exfat::Impl::Impl(const xmlNode* node)
+	: BlkFilesystem::Impl(node)
+    {
+    }
+
+
+    string
+    Exfat::Impl::get_pretty_classname() const
+    {
+	// TRANSLATORS: name of object
```
mvidner: Name of "object"?? If I'm the translator, this is useless.
What is the context? What are other examples of "objects"?

mvidner: Apparently the correct spelling is "exFAT": https://en.wikipedia.org/wiki/ExFAT

aschnell:
The tools to support exfat are included in openSUSE and all packages have to be reviewed by legal so I must assume that this is fine now. In any case, the code here does not look to be covered by any patents.

mvidner:
So the exFAT tools might be problematic and could be missing, or the user could install them from a non-opensuse repo. This code should work either way.

A .changes entry?

LGTM, except the po conflict.

aschnell:
The changelog is generated from git log.

PR #560, opened 2018-08-16
adjust multipath parser to accept nvme related output (bsc #1089353)

wfeldt:
This also fixes a bug: the old code didn't expect lines to start with ": "
resulting in mis-parsing those lines.  (The multipath element got ':"
assigned as name).


shundhammer:
LGTM

shundhammer:
LGTM

shundhammer:
LGTM

ancorgs:
We need to release this fix as MU for SLE-15 (maybe even as self-update installer, since it makes installation impossible in multipath+NVME environments). So I guess we need this in another branch (in addition to master).

PR #562, opened 2018-08-22
detect correctly whether a file system is currently mounted (bsc#1105227)

wfeldt:
This patch basically ensures that the mount point is in 'canonical' form.


shundhammer:
LGTM besides some minor whitespace inconsistencies in the test case.

PR #572, opened 2018-09-12
Allow bcache to have partitions

jreidinger:


aschnell:
Did you run utils/generate-downcast.py to generate bindings/storage-downcast.i (first run make doc)? The file storage-downcast.i may change.

Apart from that LGTM.

jreidinger:
@aschnell no, will do it. Thanks. I also have to add fate number to commit message.

jreidinger:
@aschnell btw this info is not in README, should I add it? Maybe to section about submission or compiling?

aschnell:
Sure you may improve README. But that does not block merging this PR.

PR #579, opened 2018-10-04
Use for bcache cset as display name string "bcache cache" (fate#325346)

jreidinger:
Reason is that previously it use uuid which has two drawbacks:

1. it is too long
2. it is not known when cset is not yet commited, so result is empty
string which is confusing

screenshot how it looks now ( previously there was empty box instead of "bcache caching set" )

cset_name

and to compare old look:

cset_old

### DIFF
--- a/storage/Devices/BcacheCsetImpl.h
+++ b/storage/Devices/BcacheCsetImpl.h
@@ -52,7 +52,7 @@ namespace storage
 
 	virtual const char* get_classname() const override { return DeviceTraits<BcacheCset>::classname; }
 
-	virtual string get_displayname() const override { return get_uuid(); }
+	virtual string get_displayname() const override { return "bcache cache"; }
 
 	virtual string get_pretty_classname() const override;
 


aschnell:
AFAIK the thing is usually called "cache set" not "caching set".

jreidinger:
@aschnell fine for me renaming. I even consider using just "bcache cache" as from arrows it should be visible if it is single or multiple devices. So what do you think about "bcache cache" ?

jreidinger:
another option can be just short "cache set" and not "bcache cache set"

aschnell:
The "bcache cache set" can be disconnected from a "bcache" (or is that not possible?). In that case just naming it "cache set" might not make it so clear, e.g. LVM could also have "cache set/devices".

jreidinger:
@aschnell yes, there can be bcache cache set that it not attached to any bcache device and in fact that can be block device defined on this bcache_cset ( which is like persistent cache without backing device ) which is called flash_only_volume. ( I have problem to display it right now, as I have issues with kernel/udev assembly of such flash_only_volume ). So what about that "bcache cache" ? I think that "set" keyword is just not needed as arrows should make it obvious if it is single or multiple devices and currently I think it is even not supported by suse and using md raid 1 as caching device is recommended way instead of multiple devices in caching set.

aschnell:
"bcache cache" is also fine.

jreidinger:
@aschnell thanks, so I will rename it.

jreidinger:
thanks

PR #580, opened 2018-10-09
[RFC] add bcache attributes (fate#325346)

jreidinger:
Do not merge before real world testing.

- [x] testsuite
- [x] write

### DIFF
--- a/storage/Devicegraph.cc
+++ b/storage/Devicegraph.cc
@@ -29,6 +29,7 @@
 #include "storage/Devicegraph.h"
 #include "storage/Utils/GraphUtils.h"
 #include "storage/Devices/DeviceImpl.h"
+#include "storage/Devices/Bcache.h"
 #include "storage/Devices/BlkDevice.h"
 #include "storage/Devices/Disk.h"
 #include "storage/Devices/Md.h"
@@ -339,8 +340,8 @@ namespace storage
     void
     Devicegraph::remove_device(sid_t sid)
     {
-	const Device* device = find_device(sid);
-	get_impl().remove_vertex(device->get_impl().get_vertex());
+	Device* device = find_device(sid);
+	remove_device(device);
     }
 
 
@@ -348,6 +349,8 @@ namespace storage
     Devicegraph::remove_device(Device* device)
     {
 	get_impl().remove_vertex(device->get_impl().get_vertex());
+        // ensure that bcache numbers are correct
+	Bcache::reassign_numbers(this);
     }
 
 
@@ -356,6 +359,8 @@ namespace storage
     {
 	for (Device* device : devices)
 	    get_impl().remove_vertex(device->get_impl().get_vertex());
+	// ensure that bcache numbers are correct
+	Bcache::reassign_numbers(this);
     }
 
 
--- a/storage/Devices/Bcache.cc
+++ b/storage/Devices/Bcache.cc
@@ -110,6 +110,48 @@ namespace storage
     }
 
 
+    CacheMode
+    Bcache::get_cache_mode() const
+    {
+	return get_impl().get_cache_mode();
+    }
+
+
+    void
+    Bcache::set_cache_mode(CacheMode mode)
+    {
+	get_impl().set_cache_mode(mode);
+    }
+
+
+    unsigned long long
+    Bcache::get_sequential_cutoff() const
+    {
+	return get_impl().get_sequential_cutoff();
+    }
+
+
+    void
+    Bcache::set_sequential_cutoff(unsigned long long size)
+    {
+	get_impl().set_sequential_cutoff(size);
+    }
+
+
+    unsigned
+    Bcache::get_writeback_percent() const
+    {
+	return get_impl().get_writeback_percent();
+    }
+
+
+    void
+    Bcache::set_writeback_percent(unsigned percent)
+    {
+	get_impl().set_writeback_percent(percent);
+    }
+
+
     vector<Bcache*>
     Bcache::get_all(Devicegraph* devicegraph)
     {
@@ -121,6 +163,13 @@ namespace storage
     Bcache::get_all(const Devicegraph* devicegraph)
     {
 	return devicegraph->get_impl().get_devices_of_type<const Bcache>();
+    }
+
+
+    void
+    Bcache::reassign_numbers(Devicegraph* devicegraph)
+    {
+	return Bcache::Impl::reassign_numbers(devicegraph);
     }
 
 
--- a/storage/Devices/Bcache.h
+++ b/storage/Devices/Bcache.h
@@ -34,6 +34,14 @@ namespace storage
 
 
     /**
+     * The Cache mode attribute.
+     */
+    enum class CacheMode {
+	WRITETHROUGH, WRITEBACK, WRITEAROUND, NONE
+    };
+
+
+    /**
      * A <a href="https://www.kernel.org/doc/Documentation/bcache.txt">bcache</a>
      * device.
      */
@@ -69,6 +77,42 @@ namespace storage
 	void attach_bcache_cset(BcacheCset* bcache_cset);
 
 	/**
+	 * Returns cache mode attribute.
+	 */
+	CacheMode get_cache_mode() const;
+
+	/**
+	 * Sets cache mode attribute
+	 *
+	 * @param[in] mode target cache mode
+	 */
+	void set_cache_mode(CacheMode mode);
+
+	/**
+	 * Returns size of sequential_cutoff attribute.
+	 */
+	unsigned long long get_sequential_cutoff() const;
+
+	/**
+	 * Sets sequential_cutoff attribute
+	 *
+	 * @param[in] size size in a bytes
+	 */
+	void set_sequential_cutoff(unsigned long long size);
+
+	/**
+	 * Returns percent of writeback dirty pages.
+	 */
+	unsigned get_writeback_percent() const;
+
+	/**
+	 * Sets writeback percent attribute
+	 *
+	 * @param[in] percent target cache mode
+	 */
+	void set_writeback_percent(unsigned percent);
+
+	/**
 	 * Get all Bcaches.
 	 */
 	static std::vector<Bcache*> get_all(Devicegraph* devicegraph);
@@ -77,6 +121,18 @@ namespace storage
 	 * @copydoc get_all()
 	 */
 	static std::vector<const Bcache*> get_all(const Devicegraph* devicegraph);
+
+	/**
+	 * Fix the numeric ids of bcache devices so that there are no holes in the id sequence
+	 * for in-memory bcache devices. Called automatically when a device is deleted and
+	 * a hole can appear.
+	 *
+	 * Following rules apply to bcaches:
+	 * - An existing bcache keeps its number until reboot
+	 * - A newly created device always uses the lowest available number
+	 *   So the devices in memory always fill holes in the numbering sequence
+	 */
+	static void reassign_numbers(Devicegraph* devicegraph);
 
 	/**
 	 * Find a Bcache by its name.
--- a/storage/Devices/BcacheImpl.cc
+++ b/storage/Devices/BcacheImpl.cc
@@ -24,6 +24,7 @@
 
 #include "storage/Utils/Algorithm.h"
 #include "storage/Utils/AppUtil.h"
+#include "storage/Utils/LoggerImpl.h"
 #include "storage/Utils/StorageDefines.h"
 #include "storage/Utils/XmlFile.h"
 #include "storage/Utils/SystemCmd.h"
@@ -45,25 +46,36 @@ namespace storage
     using namespace std;
 
 
-    // TODO rename after delete (in memory)
     // TODO handle attach and detach
     // TODO synchronization after actions
-    // TODO cache-mode
+    // TODO write sequential_cutoff and writeback_percent persistently to survive reboot
 
 
     const char* DeviceTraits<Bcache>::classname = "Bcache";
 
+    const vector<string> EnumTraits<CacheMode>::names({
+	"writethrough", "writeback", "writearound", "none"
+    });
 
     Bcache::Impl::Impl(const string& name)
-	: Partitionable::Impl(name)
+	: Partitionable::Impl(name), cache_mode(CacheMode::WRITETHROUGH),
+          sequential_cutoff(8 * MiB), writeback_percent(10) // bcache defaults
     {
 	update_sysfs_name_and_path();
     }
 
 
     Bcache::Impl::Impl(const xmlNode* node)
-	: Partitionable::Impl(node)
+	: Partitionable::Impl(node), cache_mode(CacheMode::WRITETHROUGH),
+          sequential_cutoff(8 * MiB), writeback_percent(10) // bcache defaults
     {
+	string tmp;
+
+	cache_mode = CacheMode::WRITETHROUGH;
+	if (getChildValue(node, "cache-mode", tmp))
+	    cache_mode = toValueWithFallback(tmp, CacheMode::WRITETHROUGH);
+	getChildValue(node, "writeback-percent", writeback_percent);
+	getChildValue(node, "sequential-cutoff", sequential_cutoff);
     }
 
 
@@ -86,6 +98,34 @@ namespace storage
     }
 
 
+    void
+    Bcache::Impl::reassign_numbers(Devicegraph* devicegraph)
+    {
+	// not many bcache devices expected (<10), so lets just do tail-recursive
+	// algorithm. It finds free name and ensure that all devices
+	// with higher number are not in memory. And if there is a memory one,
+	// then rename it to free name and run again. It always terminate as each step remove one
+	// in-memory bcache that has hole before till all in-memory caches occupy the lowest
+	// available numbers.
+	const string free_name = find_free_name(devicegraph);
+	unsigned free_number = device_to_name_and_number(free_name).second;
+
+	vector<Bcache*> bcaches = get_all(devicegraph);
+
+	sort(bcaches.begin(), bcaches.end(), compare_by_number);
+
+	for(Bcache* bcache: bcaches)
+	{
+	    if (bcache->get_number() > free_number && !bcache->exists_in_probed())
+	    {
+		bcache->get_impl().set_number(free_number);
+		reassign_numbers(devicegraph);
+		return;
+	    }
+	}
+    }
+
+
     string
     Bcache::Impl::find_free_name(const Devicegraph* devicegraph)
     {
@@ -103,6 +143,10 @@ namespace storage
     Bcache::Impl::save(xmlNode* node) const
     {
 	Partitionable::Impl::save(node);
+
+	setChildValue(node, "cache-mode", toString(cache_mode));
+	setChildValue(node, "writeback-percent", writeback_percent);
+	setChildValue(node, "sequential-cutoff", sequential_cutoff);
     }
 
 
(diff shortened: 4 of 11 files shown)


storage/Devices/BcacheImpl.cc
```diff
@@ -144,6 +145,61 @@ namespace storage
     }
 
 
+    static const map<std::string, CacheMode> cache_mode_mapping = {
```
aschnell: For such mapping there is EnumTraits, e.g. used for the FsType enum.

storage/Devices/BcacheImpl.cc
```diff
@@ -232,6 +297,48 @@ namespace storage
     }
 
 
+    CacheMode
+    Bcache::Impl::get_cache_mode() const
+    {
+        return cache_mode;
+    }
```
aschnell: Such simple functions can go in the header file and be inlined.

storage/Devices/BcacheImpl.cc
jreidinger: thanks, will check it.

storage/Devices/BcacheImpl.cc
jreidinger: Ok, makes sense. I expect that explicit `inline` keyword is no longer needed with modern compilers?

aschnell: AFAIK it was never required for functions implemented directly in the class definition and for functions implemented outside always was.

storage/Utils/StorageTmpl.h
```diff
@@ -127,7 +127,7 @@ namespace storage
 	static_assert(!std::is_enum<Type>::value, "is enum");
 
 	if (lhs != rhs)
-	    log << " " << text << ":" << lhs << "-->" << rhs;
+	    log << " " << text << ":" << lhs << "-->" << rhs << "\n";
```
jreidinger: change in this file really helps with readability when testsuite failed, as previously lines was mixed together.

Now it returns something like:

```
bcache1.cc(38): error: in "probe": 
sid 52 device differ
 cache_mode:writearound-->writethrough
 writeback_percent:0-->10
 sequential_cutoff:1024-->0
```

storage/Devices/BcacheImpl.cc
```diff
@@ -53,17 +54,29 @@ namespace storage
 
     const char* DeviceTraits<Bcache>::classname = "Bcache";
 
+    const vector<string> EnumTraits<CacheMode>::names({
```
mvidner: Since this was new to you, would it help future contributors to add a doc comment to `Enum.h` how to use `EnumTraits`?

storage/Devices/BcacheImpl.cc
```diff
@@ -53,17 +54,29 @@ namespace storage
```
mvidner: Do the TODOs above still apply?

storage/Devices/Bcache.h
```diff
@@ -33,6 +33,14 @@ namespace storage
     class BcacheCset;
 
 
+    /**
+     * The Cache mode attribute.
+     */
+    enum class CacheMode {
```
mvidner: IMHO this is specific to bcache so it should be moved inside the Bcache namespace/class.

storage/Devices/BcacheImpl.cc
```diff
@@ -53,17 +54,29 @@ namespace storage
...
+	: Partitionable::Impl(node), cache_mode(CacheMode::WRITETHROUGH),
+          sequential_cutoff(8 * MiB), writeback_percent(10)
     {
+        string tmp;
+
+        cache_mode = CacheMode::WRITETHROUGH;
+        if (getChildValue(node, "cache_mode", tmp))
+            cache_mode = toValueWithFallback(tmp, CacheMode::WRITETHROUGH);
+	getChildValue(node, "writeback_percent", writeback_percent);
+	getChildValue(node, "sequential_cutoff", sequential_cutoff);
```
mvidner: misaligned

(Which editor do you use? I collect tips to configure editors to avoid these annoyances.)

storage/Devices/BcacheImpl.cc
```diff
@@ -53,17 +54,29 @@ namespace storage
...
     const char* DeviceTraits<Bcache>::classname = "Bcache";
 
+    const vector<string> EnumTraits<CacheMode>::names({
+	"writethrough", "writeback", "writearound", "none"
+    });
 
     Bcache::Impl::Impl(const string& name)
-	: Partitionable::Impl(name)
+	: Partitionable::Impl(name), cache_mode(CacheMode::WRITETHROUGH),
+          sequential_cutoff(8 * MiB), writeback_percent(10)
```
mvidner: Where do the defaults come from?

storage/Devices/BcacheImpl.cc
```diff
@@ -144,6 +161,54 @@ namespace storage
...
+        regex rgx("\\[(.*)\\]");
+        smatch match;
+        if ( regex_search(modes, match, rgx) )
+        {
+
+            return toValueWithFallback(match[1], CacheMode::WRITETHROUGH);
+        }
+        else
+        {
+            y2err("Unknown value in cache_mode '" << modes << "'");
```
mvidner: This is misleading. This occurs for an "unparseable file", while an unknown value, like "writefancypants" is *silently* converted by `toValueWithFallback` above, right?

storage/Devices/BcacheImpl.cc
```diff
@@ -152,6 +217,15 @@ namespace storage
 	const File size_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/size");
 
 	set_region(Region(0, size_file.get<unsigned long long>(), 512));
+
+	const File cache_mode_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/bcache/cache_mode");
+        set_cache_mode(parse_mode(cache_mode_file));
+
+	const File writeback_percent_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/bcache/writeback_percent");
+        set_writeback_percent(writeback_percent_file.get<int>());
```
mvidner: unsigned instead of int

storage/Devices/BcacheImpl.cc
```diff
@@ -415,7 +502,19 @@ namespace storage
 
 	SystemCmd cmd(cmd_line, SystemCmd::DoThrow);
 
-	sleep(1);
+	sleep(1); // TODO: is it enough? check with Coly how to recognize it
```
mvidner: WTF, please add a comment *why* a sleep is needed.

storage/Devices/BcacheImpl.cc
jreidinger: well, I think it can help. But what would help the most is to know that such class exists at all.

storage/Devices/BcacheImpl.cc
jreidinger: I think yes, except cache mode.

storage/Devices/Bcache.h
jreidinger: It is consistent with rest of library which also has it outside like MD raid level and others.

storage/Devices/BcacheImpl.cc
jreidinger: to be honest, I prefer if there is e.g. clang format configuration, so I can check it miself. I think reason why it is misaligned is that libstorage-ng use ancient yast style with mixture of tabs and spaces, which is not so easy to configure in editor (at least I never succeed ).

storage/Devices/BcacheImpl.cc
jreidinger: it is bcache defaults for new device.

storage/Devices/BcacheImpl.cc
jreidinger: yeah, result of previous alghorithm

storage/Devices/BcacheImpl.cc
mvidner: Please see https://github.com/openSUSE/libstorage-ng/blob/master/.editorconfig

storage/Devices/BcacheImpl.cc
jreidinger: no, this won't work, as get is only specialized and has no specialization for unsigned ( I originally have it ). But as long as it is percents so 0..100 we are safe even with int.

storage/Devices/BcacheImpl.cc
jreidinger: OK, reason is that all bcache sysfs operations are async, so hard to say when it ends.

storage/Devices/Bcache.h
mvidner: Uggggggh OK.

storage/Devices/BcacheImpl.cc
aschnell: Yes, that .editorconfig seems to properly represent doc/style-guide.md.

Concerning clang format the result I had so far were always unsatisfactory.

storage/Devices/BcacheImpl.cc
mvidner: WTF, then we must add a loop to wait for completion, or we will get super funny bugs, right?

aschnell: Yes, an in BcacheImpl.cc are several TODOs, one being "synchronization after actions".

storage/Devices/BcacheImpl.cc
mvidner: Well, since `File::get<int>()` contains "// TODO error checking", this is a full grown skeleton anyway.

storage/Devices/BcacheImpl.cc
mvidner: += `// these are the bcache defaults for a new device`

storage/Devices/BcacheImpl.cc
mvidner: so? hint hint

storage/Devices/BcacheImpl.cc
mvidner: I don't understand. Anyway:
- change the error message to "Cannot parse cache_mode '"
- raise an exception, both for unparseable input, and a new unknown enum value, which is consistent with the rest of the library, I HOPE.

storage/Devices/BcacheImpl.cc
jreidinger: @mvidner in theory yes, but for same actions I need coly to say me which file have to appear and also some operation together with certain state of cache result in several hours or days till finish of operation ( detach one ). I can forward you notes from my conversation with Coly when he was in PRG.

storage/Devices/BcacheImpl.cc
jreidinger: With more testing with bcache with various size I actually found that 10% is correct, but sequential cutoff is dynamic and depends "probably" on caching device size. As now for smaller disk I have default 4 MiB.

storage/Devices/BcacheImpl.cc
jreidinger: I check it and majority of library use variant with fallback. It will log error for failing case. So I keep it as it is.

storage/Devices/BcacheImpl.cc
```diff
@@ -45,25 +46,35 @@ namespace storage
...
 
     Bcache::Impl::Impl(const xmlNode* node)
-	: Partitionable::Impl(node)
+	: Partitionable::Impl(node), cache_mode(CacheMode::WRITETHROUGH),
+          sequential_cutoff(8 * MiB), writeback_percent(10) // bcache defaults
     {
+	string tmp;
+
+	cache_mode = CacheMode::WRITETHROUGH;
+	if (getChildValue(node, "cache_mode", tmp))
```
aschnell: The convention is to use '-' instead of '_' in the xml files.

storage/Devices/BcacheImpl.cc
```diff
@@ -415,7 +536,20 @@ namespace storage
...
+
+	SystemCmd cmd2(cmd_line, SystemCmd::DoThrow);
+
+	cmd_line = ECHO_BIN " " + to_string(writeback_percent) + " > " + quote(SYSFS_DIR "/" + get_sysfs_path() + "/bcache/writeback_percent");
+
+	SystemCmd cmd3(cmd_line, SystemCmd::DoThrow);
+
+	cmd_line = ECHO_BIN " " + to_string(sequential_cutoff) + " > " + quote(SYSFS_DIR "/" + get_sysfs_path() + "/bcache/sequential_cutoff");
+
+	SystemCmd cmd4(cmd_line, SystemCmd::DoThrow);
```
aschnell: Setting the attributes here is fine. But it will not allow to just change the attributes without creating the bcache. For that new actions (or one big action) are required, see e.g. Gpt::Impl::add_modify_actions().

For the time being a TODO notice is fine for me.

storage/Devices/BcacheImpl.cc
aschnell: Yes, the others still apply. But that does not block this PR.

storage/Devices/Bcache.cc
```diff
@@ -124,6 +166,13 @@ namespace storage
     }
 
 
+    void
+    Bcache::reassign_numbers(Devicegraph* devicegraph)
+    {
```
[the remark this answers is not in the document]
jreidinger: It can be only internal. I think it can be useful from outside when needed to fix numbers even when no device removal is done. E.g. when construction names from autoyast profile and if there will be some holes, then this method will fix it.

storage/Devices/BcacheImpl.cc
jreidinger: ok, will change

storage/Devices/BcacheImpl.cc
jreidinger: OK, for now it is a bit intention as with modify there is a problem with that async operations, which is not needed for cache_mode, but others will need at first detach, then modify and then attach again ( source Coly, but I found on network, then e.g. writeback_percent should be possible to modify without detach, so still something that needs to be make clear ).

So I will add that TODO.

storage/Devices/BcacheImpl.cc
```diff
@@ -152,6 +240,15 @@ namespace storage
...
 	set_region(Region(0, size_file.get<unsigned long long>(), 512));
+
+	const File cache_mode_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/bcache/cache_mode");
+	set_cache_mode(parse_mode(cache_mode_file));
+
+	const File writeback_percent_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/bcache/writeback_percent");
+	set_writeback_percent(writeback_percent_file.get<unsigned>());
+
+	const File sequential_cutoff_file = prober.get_system_info().getFile(SYSFS_DIR + get_sysfs_path() + "/bcache/sequential_cutoff");
+	set_sequential_cutoff(parse_size(sequential_cutoff_file));
```
aschnell: When ```prober.get_system_info()``` is used several times it can be saved in a reference: ```SystemInfo& system_info = prober.get_system_info();```

storage/Devices/Bcache.cc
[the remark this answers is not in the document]
aschnell: OK, that is a good reason.

storage/Devicegraph.cc
```diff
@@ -339,15 +340,17 @@ namespace storage
...
+	Device* device = find_device(sid);
+	remove_device(device);
     }
 
 
     void
     Devicegraph::remove_device(Device* device)
     {
 	get_impl().remove_vertex(device->get_impl().get_vertex());
+        // ensure that bcache numbers are correct
```
mvidner: misaligned, also below

storage/Devices/BcacheImpl.cc
mvidner: I meant to remove those that are addressed in this PR, which has been done.

storage/Devices/BcacheImpl.cc
mvidner: Fair point, EditorConfig addresses the "encoding" and "implementation" of "indents" but it does not help you with specifying where the indents should go (or braces, for that matter).

storage/Devices/BcacheImpl.cc
```diff
@@ -103,6 +140,10 @@ namespace storage
     Bcache::Impl::save(xmlNode* node) const
     {
 	Partitionable::Impl::save(node);
+
+	setChildValue(node, "cache_mode", toString(cache_mode));
+	setChildValue(node, "writeback_percent", writeback_percent);
+	setChildValue(node, "sequential_cutoff", sequential_cutoff);
```
mvidner: `_` vs `-` also here

storage/SystemInfo/DevAndSys.cc
```diff
@@ -108,14 +108,14 @@ namespace storage
...
+    File::get() const
     {
 	if (content.empty())
 	    ST_THROW(Exception("empty file " + path));
 
-	int ret;
+	T ret;
 	content.front() >> ret;
 
 	// TODO error checking
```
mvidner: Since we're touching this method...
According to 
- http://www.cplusplus.com/reference/istream/istream/operator%3E%3E/ 
- http://www.cplusplus.com/reference/ios/ios/rdstate/

we can check for an error with `if (!content.good())`.

Given the `ST_THROW` above, I think another one here would be fine.

testsuite/bcache1.cc
```diff
@@ -30,3 +30,29 @@ BOOST_AUTO_TEST_CASE(free_name)
...
+
+    // we need some bcaches on system, here just ensure that no unnoticed xml change happen.
+    BOOST_CHECK_EQUAL(Bcache::get_all(storage.get_probed()).size(), 3);
+
+    Devicegraph* staging = storage.get_staging();
+
+    const Bcache *new_bcache1 = Bcache::create(staging, "/dev/bcache3");
+    const Bcache *new_bcache2 = Bcache::create(staging, "/dev/bcache4");
+
+    // a bit hidden, but that remove_device call inside reassign numbers,
```
mvidner: > a bit hidden, but remove_device does call reassign_numbers,

storage/Devices/BcacheImpl.cc
jreidinger: ok, will modify

storage/Devices/BcacheImpl.cc
jreidinger: @mvidner but thanks for it, as it helps for lines where it does not use tabs properly and now new lines looks properly indented. So using it carefully to parts of file only and for new lines.

storage/Devices/Bcache.h
```diff
@@ -78,6 +122,16 @@ namespace storage
 	 */
 	static std::vector<const Bcache*> get_all(const Devicegraph* devicegraph);
 
+	/**
+	 * Fix numbers of bcache devices, so there is no holes in numbers between in memory
```
mvidner: > Fix the numeric ids of bcache devices so that there are no holes in the id sequence for in-memory
bcache devices. Called automatically when a device is deleted and a hole can appear.

storage/Devices/Bcache.h
```diff
@@ -78,6 +122,16 @@ namespace storage
...
+	 * bcache devices. Called automatic when device is deleted and hole can appear.
+	 *
+	 * Following rules apply to bcaches: Existing bcache devices keeps its number till reboot.
```
mvidner:
```
Following rules apply to bcaches:
- An existing bcache keeps its number until reboot
- A newly created device always uses the lowest available number
So the devices in memory always fill holes in the numbering sequence
```
("from the beginning of device", WTF, I've corrected that, OK?)

mvidner:
@jreidinger In the code we keep talking about "devices *in memory*". Let's make sure we mean the same thing: They are devices that libstorage is planning, but has not committed to the system yet. Right?

storage/Devices/BcacheImpl.cc
```diff
@@ -86,6 +97,32 @@ namespace storage
     }
 
 
+    void
+    Bcache::Impl::reassign_numbers(Devicegraph* devicegraph)
+    {
+	// not many bcache devices expected (<10), so lets just do trivial
+	// recursive algorithm. It finds free name and ensure that all devices
```
mvidner: trivial recursive :rofl: 
The important thing to mention is "tail-recursive", practically a loop, which is easy to see thanks to the `return`, but also making sure the recursion does terminate. Can we explain that it indeed does?

jreidinger:
@mvidner yes, it is devices that does not exists yet, but are planned to be created in target device graph.

storage/Devices/BcacheImpl.cc
mvidner: @jreidinger sorry, the grammar is so bad that I don't understand what you mean. can you say that again, more carefully, please?

examples: modify != modification, modifying; detach != detaching; is != are; then != than != that

storage/Devicegraph.cc
jreidinger: thanks. fixed.

storage/SystemInfo/DevAndSys.cc
jreidinger: @mvidner but content according to header file, it is vector<string> so not istream, so I worruy iostream methods won't work? I think it is converted by some boost string extension? For me searching for that operand in boost documentation does not end well ( for me boost documentation is a bit hard to read and its header files even harder :)

storage/Devices/BcacheImpl.cc
jreidinger: @mvidner OK. So basically for now we support only add/remove bcache device. Changing configuration for existing bcache device is problematic as some operations require detach of cache, which is expensive operation.

storage/SystemInfo/DevAndSys.cc
aschnell: Yes,  ```content``` is a ```vector``` not a ```stream```. Martin likely meant to use ```good()``` on a ```stream``` but in the code is no ```stream``` visible. The ```operator>>``` from StorageTmpl.h is used and hides the ```stream```. So the few lines from ```operator>>``` could be copied here and then error checking can be added,

storage/Devices/BcacheImpl.cc
aschnell: The recursive implementation does not look so efficient anyway since it queries all bcaches from the devicegraph (get_all) every time.

jreidinger:
@aschnell thanks, merge waiting for making clear how to make change to cutoff and writeback persistent, as currently it does not survive reboot. So I am waiting for Coly's answer.

jreidinger:
in the end decision is to keep the API there, add TODO for that params and do not promote it yet in YaST UI as it is not persistent. Coly do not answer yet.

PR #581, opened 2018-10-10
 Fixed blkid output parser to honor escaped quotes

shundhammer:
## Trello

https://trello.com/c/spdBxnv9/371-betasles-15-p3-1102572-probing-fails-while-parsing-a-uuid-with-protected-quote

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1102572

## Problem

When a UUID has weird characters, the _blkid_ command's output may contain escaped quotes `\"` which broke the old parser that was left over from the old libstorage.

## Solution

Wrote a new parser to split one line of the output. Since the generic `splitString()` parser from `Utils/AppUtil.cpp` is too generic to allow for that, that old one could not really be adapted while keeping it as generic as it is.

Also added quite a few unit tests for that new low-level `split_line()` parser to test fringe cases.

### DIFF
--- a/storage/SystemInfo/CmdBlkid.h
+++ b/storage/SystemInfo/CmdBlkid.h
@@ -27,6 +27,7 @@
 
 #include <string>
 #include <map>
+#include <list>
 #include <vector>
 
 #include "storage/Filesystems/Filesystem.h"
@@ -36,6 +37,7 @@ namespace storage
 {
     using std::string;
     using std::map;
+    using std::list;
     using std::vector;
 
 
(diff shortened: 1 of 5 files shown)


storage/SystemInfo/CmdBlkid.h
```diff
@@ -103,6 +105,8 @@ namespace storage
 	bool any_luks() const;
 	bool any_bcache() const;
 
+        static list<string> split_line( const string & line );
```
aschnell: In general vector is better than list. Since makeMap is only used in CmdBlkid that could be change now. Clearly optional.

aschnell:
LGTM

PR #590, opened 2018-11-08
Activate SWIG trackobject feature

joseivanlopez:
Using "trackobjects" SWIG feature to avoid aleatory unit tests crashes at IBS and OBS, for example: https://build.suse.de/build/Devel:YaST:Head/SUSE_SLE-15-SP1_GA/s390x/yast2-storage-ng/_log

Maybe this does not solve the problem at all, only giving it a try.

### DIFF
--- a/bindings/ruby/storage-ruby.i
+++ b/bindings/ruby/storage-ruby.i
@@ -37,3 +37,6 @@
 
 %include "../storage.i"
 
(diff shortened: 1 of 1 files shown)


joseivanlopez:
Actually this is not needed to fix the yast-storage-ng test suite, see https://github.com/yast/yast-storage-ng/pull/795

PR #593, opened 2018-11-13
- added support to create udf

aschnell:


storage/Filesystems/UdfImpl.cc
```diff
@@ -39,11 +43,51 @@ namespace storage
...
+	return UF_UDF | BlkFilesystem::Impl::used_features();
+    }
+
+
+    void
+    Udf::Impl::do_create()
+    {
+	const BlkDevice* blk_device = get_blk_device();
+
+	string cmd_line = MKFS_UDF_BIN " " + get_mkfs_options() + " --label " + quote(get_label()) +
```
shundhammer: Does this work if the filesystem has no label? Or will the mkfs command throw an error then?

shundhammer:
Other than the one question I asked, LGTM.

storage/Filesystems/UdfImpl.cc
aschnell: It works and is even required since otherwise mkfs.udf sets the label to "LinuxUDF".

PR #595, opened 2018-11-23
- improved sformat to be type safe and handle Text properly

aschnell:
For https://trello.com/c/TOPKOtFk/18-replace-the-sprintf-valist-based-sformat-with-something-typesafe.


mlandres:
lgtm

aschnell:
Thanks.

PR #597, opened 2018-11-23
Fix audit issues

joseivanlopez:
PBI: https://trello.com/c/qdxDTeL9/372-p3-1059972-libstorage-ng-audit-short-term-fixes

Note: the refactorization of SystemCmd will be addressed via FATE (too huge change for this phase).

### DIFF
--- a/storage/EtcCrypttab.cc
+++ b/storage/EtcCrypttab.cc
@@ -157,8 +157,8 @@ namespace storage
     }
 
 
-    EtcCrypttab::EtcCrypttab(const string& filename) :
-	ColumnConfigFile()
+    EtcCrypttab::EtcCrypttab(const string& filename, int permissions) :
+	ColumnConfigFile(permissions)
     {
 	// Set reasonable field widths for /etc/crypttab
 
--- a/storage/EtcCrypttab.h
+++ b/storage/EtcCrypttab.h
@@ -202,7 +202,13 @@ namespace storage
 
     private:
 
-	static const int DEFAULT_PERMISSIONS = 0640;
+	/**
+	 * libstorage-ng does not write cleartext passwords in crypttab file,
+	 * but other tools or admins might add cleartext passwords in there.
+	 * Therefore, this file should not be world-readable, otherwise it might
+	 * expose cleartext passwords to unprivileged users.
+	 */
+	static const int DEFAULT_PERMISSIONS = 0600;
     };
 
 }
--- a/storage/EtcCrypttab.h
+++ b/storage/EtcCrypttab.h
@@ -139,7 +139,7 @@ namespace storage
     {
     public:
 
-	EtcCrypttab(const string& filename = ETC_CRYPTTAB);
+	EtcCrypttab(const string& filename = ETC_CRYPTTAB, int permissions = DEFAULT_PERMISSIONS);
 	virtual ~EtcCrypttab();
 
 	// using inherited read() and write() unchanged
@@ -199,6 +199,10 @@ namespace storage
          * Dump the current contents to the log.
          **/
         void log();
+
+    private:
+
+	static const int DEFAULT_PERMISSIONS = 0640;
     };
 
 }
--- a/storage/Filesystems/BtrfsSubvolumeImpl.cc
+++ b/storage/Filesystems/BtrfsSubvolumeImpl.cc
@@ -193,6 +193,11 @@ namespace storage
     MountOpts
     BtrfsSubvolume::Impl::get_default_mount_options() const
     {
+	// NOTE: This is used for both: fstab and mount command.
+	// In case of fstab, subvol path cannot be quoted, and the usage of scape
+	// commands like \040 and \011 is undefined (and probably not supported by
+	// the tools). Comma might be dangerous due to it is used as separator for
+	// mount options.
 	return MountOpts(vector<string>({ "subvol=/" + path }));
     }
 
@@ -200,6 +205,7 @@ namespace storage
     vector<string>
     BtrfsSubvolume::Impl::get_mount_options() const
     {
+	// See get_default_mount_options()
 	return vector<string>({ "subvol=/" + path });
     }
 
--- a/storage/Utils/AsciiFile.cc
+++ b/storage/Utils/AsciiFile.cc
@@ -120,7 +120,12 @@ namespace storage
 	{
 	    y2mil("saving file " << name);
 
-	    FILE* file = fdopen(open(name.c_str(), O_WRONLY | O_TRUNC | O_CREAT, permissions), "w");
+	    int fd = open(name.c_str(), O_WRONLY | O_TRUNC | O_CREAT | O_CLOEXEC, permissions);
+
+	    if (fd < 0)
+		ST_THROW(IOException(sformat("Opening file %s failed: %s", name, strerror(errno))));
+
+	    FILE* file = fdopen(fd, "we");
 
 	    if (!file)
 		ST_THROW(IOException(sformat("Opening file %s failed: %s", name, strerror(errno))));
@@ -130,13 +135,13 @@ namespace storage
 		string line_with_break = line + "\n";
 
 		fputs(line_with_break.c_str(), file);
-
-		if (ferror(file))
-		    ST_THROW(IOException(sformat("Saving file %s failed: %s", name, strerror(errno))));
 	    }
 
-	    if(fclose(file) == EOF)
-		y2war("closing file " << name << "failed");
+	    if (ferror(file))
+		ST_THROW(IOException(sformat("Saving file %s failed: %s", name, strerror(errno))));
+
+	    if (fclose(file) != 0)
+		ST_THROW(IOException(sformat("Closing file %s failed: %s", name, strerror(errno))));
 	}
     }
 
--- a/storage/Utils/AsciiFile.cc
+++ b/storage/Utils/AsciiFile.cc
@@ -23,12 +23,18 @@
 
 #include <unistd.h>
 #include <fstream>
+#include <stdio.h>
+#include <fcntl.h>
+/* Not technically required, but needed on some UNIX distributions */
+#include <sys/types.h>
+#include <sys/stat.h>
 
 #include "storage/Utils/LoggerImpl.h"
 #include "storage/Utils/AsciiFile.h"
 #include "storage/Utils/Mockup.h"
 #include "storage/Utils/StorageTypes.h"
 #include "storage/Utils/Format.h"
+#include "storage/Utils/ExceptionImpl.h"
 
 
 namespace storage
@@ -36,8 +42,8 @@ namespace storage
     using namespace std;
 
 
-    AsciiFile::AsciiFile(const string& name, bool remove_empty)
-	: name(name), remove_empty(remove_empty)
+    AsciiFile::AsciiFile(const string& name, bool remove_empty, int permissions)
+	: name(name), remove_empty(remove_empty), permissions(permissions)
     {
 	reload();
     }
@@ -114,16 +120,23 @@ namespace storage
 	{
 	    y2mil("saving file " << name);
 
-	    ofstream file(name);
-	    classic(file);
+	    FILE* file = fdopen(open(name.c_str(), O_WRONLY | O_TRUNC | O_CREAT, permissions), "w");
+
+	    if (!file)
+		ST_THROW(IOException(sformat("Opening file %s failed: %s", name, strerror(errno))));
 
 	    for (const string& line : lines)
-		file << line << '\n';
+	    {
+		string line_with_break = line + "\n";
 
-	    file.close();
+		fputs(line_with_break.c_str(), file);
 
-	    if (!file.good())
-		ST_THROW(IOException(sformat("Saving file %s failed.", name)));
+		if (ferror(file))
+		    ST_THROW(IOException(sformat("Saving file %s failed: %s", name, strerror(errno))));
+	    }
+
+	    if(fclose(file) == EOF)
+		y2war("closing file " << name << "failed");
 	}
     }
 
--- a/storage/Utils/AsciiFile.h
+++ b/storage/Utils/AsciiFile.h
@@ -39,11 +39,15 @@ namespace storage
     {
     public:
 
-	explicit AsciiFile(const string& name, bool remove_empty = false);
+	explicit AsciiFile(const string& name, bool remove_empty = false, int permissions = DEFAULT_PERMISSIONS);
 
 	const string& get_name() const { return name; }
 
 	bool reload();
+
+	/**
+	 * @throw IOException
+	 */
 	void save();
 
 	void log_content() const;
@@ -55,10 +59,15 @@ namespace storage
 	vector<string>& get_lines() { return lines; }
 	const vector<string>& get_lines() const { return lines; }
 
+	void set_lines(const vector<string>& lines) { this->lines = lines; }
+
     protected:
+
+	static const int DEFAULT_PERMISSIONS = 0666;
 
 	const string name;
 	const bool remove_empty;
+	const int permissions;
 
 	vector<string> lines;
 
--- a/storage/Utils/AsciiFile.h
+++ b/storage/Utils/AsciiFile.h
@@ -63,6 +63,7 @@ namespace storage
 
     protected:
 
+	// By default, file permissions are only limited by umask value
 	static const int DEFAULT_PERMISSIONS = 0666;
 
 	const string name;
--- a/storage/Utils/ColumnConfigFile.h
+++ b/storage/Utils/ColumnConfigFile.h
@@ -117,7 +117,7 @@ public:
     };
 
 
-    ColumnConfigFile();
+    ColumnConfigFile(int permissions = DEFAULT_PERMISSIONS);
     virtual ~ColumnConfigFile();
 
     /**
@@ -187,6 +187,10 @@ protected:
     vector<int> max_column_widths;
     int         max_column_width;
     bool        pad_columns;
+
+private:
+
+    static const int DEFAULT_PERMISSIONS = 0666;
 };
 
 } // namespace storage
--- a/storage/Utils/ColumnConfigFile.h
+++ b/storage/Utils/ColumnConfigFile.h
@@ -190,6 +190,7 @@ protected:
 
 private:
 
+    // By default, file permissions are only limited by umask value
     static const int DEFAULT_PERMISSIONS = 0666;
 };
 
--- a/storage/Utils/CommentedConfigFile.cc
+++ b/storage/Utils/CommentedConfigFile.cc
@@ -20,7 +20,6 @@
  */
 
 
-#include <fstream>
 #include <algorithm>
 #include <boost/algorithm/string.hpp>
 
@@ -38,7 +37,8 @@
 using namespace storage;
 
 
-CommentedConfigFile::CommentedConfigFile():
+CommentedConfigFile::CommentedConfigFile(int permissions) :
+    permissions(permissions),
     comment_marker( "#" ),
     diff_enabled( false )
 {
@@ -138,30 +138,25 @@ void CommentedConfigFile::write( const string & new_filename )
     string name = new_filename;
 
     if ( new_filename.empty() )
-        name = this->filename;
+	name = this->filename;
     else
-        this->filename = name;
+	this->filename = name;
 
     if ( name.empty() ) // Support for mocking:
-        return;         // Pretend everything worked just fine.
-
-    std::ofstream file( name, std::ofstream::out | std::ofstream::trunc );
-
-    if ( ! file.is_open() )
-	ST_THROW(IOException(sformat("Saving file %s failed.", name)));
+	return;		// Pretend everything worked just fine.
 
     string_vec lines = format_lines();
 
-    for ( size_t i=0; i < lines.size(); ++i )
-        file << lines[i] << "\n"; // no endl: Don't flush after every line
-
     if ( diff_enabled )
-        save_orig( lines );
+	save_orig( lines );
 
-    file.close();
+    AsciiFile ascii_file(filename,
+			 true, // removes the existing file if it is empty
+			 permissions);
 
-    if ( !file.good() )
-	ST_THROW(IOException(sformat("Saving file %s failed.", name)));
+    ascii_file.set_lines(lines);
+
+    ascii_file.save();
 }
 
 
--- a/storage/Utils/CommentedConfigFile.h
+++ b/storage/Utils/CommentedConfigFile.h
@@ -520,6 +520,7 @@ protected:
 
 private:
 
+    // By default, file permissions are only limited by umask value
     static const int DEFAULT_PERMISSIONS = 0666;
 
     string	    filename;
--- a/storage/Utils/CommentedConfigFile.h
+++ b/storage/Utils/CommentedConfigFile.h
@@ -242,7 +242,7 @@ public:
      * right away because that would call virtual methods indirectly that
      * cannot be called in the constructor.
      **/
-    CommentedConfigFile();
+    CommentedConfigFile(int permissions = DEFAULT_PERMISSIONS);
 
     /**
      * Destructor. This deletes all entries.
@@ -261,7 +261,7 @@ public:
      * original file that was used in the constructor or during the last
      * read().
      *
-     * Throws an IOException if writing fails.
+     * @throw IOException
      **/
     void write( const string & filename = "" );
 
@@ -520,9 +520,12 @@ protected:
 
 private:
 
+    static const int DEFAULT_PERMISSIONS = 0666;
+
     string	    filename;
+    int 	    permissions;
     string	    comment_marker;
-    bool            diff_enabled;
+    bool	    diff_enabled;
 
     string_vec	    header_comments;
     vector<Entry *> entries;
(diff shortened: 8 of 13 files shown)


aschnell:
LGTM

aschnell:
You should mention the bug number in the commit message.

storage/Utils/LockImpl.cc
```diff
@@ -66,8 +66,10 @@ namespace storage
 		y2err("creating directory for lock-file failed: " << strerror(errno));
 	    }
 
-	    fd = open(LOCK_DIR "/lock", (read_only ? O_RDONLY : O_WRONLY) | O_CREAT | O_CLOEXEC,
-		      S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
+	    const int permissions = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH; // 0644
```
shundhammer: Personally I find the octal value 0644 **much** better to read. Why use those constants?

storage/Utils/LockImpl.cc
```diff
@@ -66,8 +66,10 @@ namespace storage
...
+
+	    fd = open(LOCK_DIR "/lock", (read_only ? O_RDONLY : O_WRONLY) | O_CREAT | O_CLOEXEC, permissions);
```
shundhammer: Since you already have a variable for those permissions, why change them here again?
Why not work on that _permissions_ variable before this line and then just pass _permissions_ to this _open()_ call? This would be much better to read and much better maintainable.

storage/Utils/LockImpl.cc
joseivanlopez: In fact, I had it with octal numbers, but not sure if octal value and constants will remain the same in the future. So, I came with the "safest" version.

examples/SystemInfo/test-cmd-btrfs-filesystem-show.cc
```diff
@@ -26,7 +26,7 @@ test_cmd_btrfs_filesystem_show(SystemInfo& system_info)
 int
 main()
 {
-    set_logger(get_logfile_logger());
+    Logger::create("/var/log/libstorage-ng.log");
```
shundhammer: I don't like the amout of copy&paste that is going on here. I also don't like this vast amount of string literals scattered all over the test suite. This clearly needs to be unified.

storage/EtcCrypttab.cc
```diff
@@ -157,8 +157,10 @@ namespace storage
     }
 
 
-    EtcCrypttab::EtcCrypttab(const string& filename) :
-	ColumnConfigFile()
+    const int EtcCrypttab::DEFAULT_PERMISSIONS = S_IRUSR | S_IWUSR | S_IRGRP; // 0640
```
shundhammer: Again, since the comment is the one thing that makes this readable for developers, why not simply use octal 0640 here? That would be much better readable and maintainable.

storage/EtcCrypttab.h
```diff
@@ -199,9 +199,12 @@ namespace storage
          * Dump the current contents to the log.
          **/
         void log();
+
+    private:
+
+	static const int DEFAULT_PERMISSIONS;
```
shundhammer: Having this const in two separate places makes this even harder to read.

This **clearly** belongs very near to the function prototype where it is used. Better yet, don't make this over-complicated; forget that constant and just use the literal value in the function prototype.

storage/Utils/LockImpl.cc
joseivanlopez: `open()` is called with three parameters, and permissions is the last one. The second parameter is the `oflags`, and I have left it untouched.

storage/Utils/LockImpl.cc
shundhammer: If this ever changes, we are doomed.

Unix permissions have relied on this since the early 1970. This is not anything that will possibly _ever_ change.

Seriously, **nobody** uses those constants, and with the cryptic letters in them, they are a lot harder to read than the octal numbers.

storage/Utils/LockImpl.cc
shundhammer: OK. Silly me.

storage/Utils/LockImpl.cc
joseivanlopez: Ok, I will change them.

examples/SystemInfo/test-cmd-btrfs-filesystem-show.cc
joseivanlopez: One of the suggested security improvements was to avoid "hard coded path" for the logs. So we have adapted the Logger to always receive the path of the log file. Anyway, I agree that the testing log path should be defined in only one place. I will fix that.

storage/EtcCrypttab.cc
```diff
@@ -157,8 +156,10 @@ namespace storage
     }
 
 
-    EtcCrypttab::EtcCrypttab(const string& filename) :
-	ColumnConfigFile()
+    const int EtcCrypttab::DEFAULT_PERMISSIONS = 0640;
+
+    EtcCrypttab::EtcCrypttab(const string& filename, const int permissions) :
```
shundhammer: `const` for function parameters only makes sense if it's a const reference, not for basic types like _int_. Please remove the `const` here.

storage/EtcCrypttab.h
```diff
@@ -139,7 +139,7 @@ namespace storage
     {
     public:
 
-	EtcCrypttab(const string& filename = ETC_CRYPTTAB);
+	EtcCrypttab(const string& filename = ETC_CRYPTTAB, const int permissions = DEFAULT_PERMISSIONS);
```
shundhammer: Same as above: Please remove `const` from the _int_ parameter.

storage/Utils/FileUtils.cc
```diff
@@ -37,6 +43,23 @@
 namespace storage
 {
 
+    bool create_file_if_not_exists(const string& filename, const int permissions)
```
shundhammer: Bad English. Please rename to `create_file_if_not_existing`.

storage/Utils/FileUtils.cc
```diff
@@ -37,6 +43,23 @@
...
+    {
+	int file_descriptor = open(filename.c_str(), O_RDWR | O_APPEND | O_CREAT, permissions);
+
+	if (file_descriptor < 0)
+	{
+	    y2err("open(" << filename << ") failed");
```
shundhammer: Please also log `errno` and `strerror(errno)` so we have a chance to find out **why** it failed.

storage/Utils/Logger.cc
```diff
@@ -20,53 +20,49 @@
  * find current contact information at www.novell.com.
  */
 
-
+#include <fcntl.h>
```
shundhammer: Do we really need that header file? This is so low-level I can't remember when I last needed it.

joseivanlopez: Leftover, I added it for the constants that we are replaced by octal numbers.

storage/Utils/Logger.cc
```diff
@@ -155,7 +132,24 @@ namespace storage
 	--count;
     }
 
-
     int Silencer::count = 0;
 
+
+    const string default_log_path()
+    {
+	static const string default_path = "/var/log/libstorage-ng.log";
```
shundhammer: This is a typical one-liner. There is no need to declare the constant when all you ever do with it is return it in the next line of code. Please simply _return_ that string right here and now. Yes, it is a `const char []` technically, but since there is a std::string constructor for that, it will automatically be converted to a std::string.

storage/Utils/Logger.cc
```diff
@@ -155,7 +132,24 @@ namespace storage
 	--count;
     }
 
-
     int Silencer::count = 0;
 
+
+    const string default_log_path()
```
shundhammer: Please use either a plain `string` or a `const string &`. Just `const string` does not serve any purpose.

storage/Utils/Logger.h
```diff
@@ -20,29 +20,28 @@
...
     /**
-     * The Logger class.
+     * Base Logger class
      */
     class Logger
     {
     public:
 
-	Logger() {}
+	static Logger* create(const std::string& filename = "");
```
shundhammer: So, is this a singleton class or not?

It looks weird to me to have both a `create()` factory method and a singleton `get_instance()` method. Which one is it? I am confused.

shundhammer: It is also unclear who owns the created instance; that should be documented here. If the caller owns it, is he supposed to delete it? Is the pointer stored in the class? Is it safe to be deleted? It the instance intentionally leaked? What happens if this `create()` is called multiple times? Is the old instance deleted in that case, or is the old instance leaked?

If this is only useful for the test suite, maybe it should rather be called someting like `create_testsuite_logger()`?

storage/EtcCrypttab.cc
shundhammer: It is not only overkill, it is also against all C++ conventions. :smile: 

Any modification to the parameter will not be visible to the caller anyway because it's call by value.

storage/Filesystems/BtrfsSubvolumeImpl.cc
```diff
@@ -193,14 +193,14 @@ namespace storage
     MountOpts
     BtrfsSubvolume::Impl::get_default_mount_options() const
     {
-	return MountOpts(vector<string>({ "subvol=/" + path }));
+	return MountOpts(vector<string>({ "subvol=" + quote("/" + path) }));
```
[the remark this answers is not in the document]
shundhammer: Duh. You are right. Silly me.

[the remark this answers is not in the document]
shundhammer: Oops - no; it's even worse.

Since this is about mount options, you cannot simply use single or double quotes; the fstab format has a very specific way of quoting; see _man fstab_. The EtcFstab class already handles this. If you add quotes here, this will result in a syntax error.

So this **needs** to remain without quotes.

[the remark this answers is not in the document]
shundhammer: https://github.com/openSUSE/libstorage-ng/blob/master/storage/EtcFstab.cc#L242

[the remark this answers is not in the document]
shundhammer: It turns out that it's unspecified how this should be handled; _man fstab_ only mentions the device and the mount point field and how to use `\040` for a space there. I could not find yet any specification for those subvolume names.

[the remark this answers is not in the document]
shundhammer: _man 5 btrfs_ also only says

> subvol=path
>
> Mount subvolume from path rather than the toplevel subvolume. The path is always treated as relative to the toplevel subvolume. This mount option overrides the default subvolume set for the given filesystem.

No mention how to handle special characters.

But if we allow a blank here, even if it is inside single or double quotes, it is clearly a syntax error for the fstab file format.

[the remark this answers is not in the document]
shundhammer: For the time being, please use `EtcFstab::fstab_encode()` instead of this `quote()`. This is the only safe way that I can see.

If there are no blanks in the path, there is no harm done. If there are, at least it won't cause a syntax error for parsing the entire fstab and thus a complete DoS for booting that system.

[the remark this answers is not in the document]
shundhammer: Please notice that it will still fail if somebody uses a comma which is the separator for mount options. I don't think fstab can handle that at all.

aschnell:
Please avoid reordering functions/classes, e.g. in Logger.h. It makes the diff bigger and less readable.

storage/Filesystems/BtrfsSubvolumeImpl.cc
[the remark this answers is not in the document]
joseivanlopez: I was playing with it, and I see that it works with command `mount` but not as fstab entry:

* This works: `mount subvol='/@/tmp' /tmp`
* But this fstbab entry does not work: `UUID=121213 /tmp btrfs subvol='/@/tmp'`

So, as you suggested, I will use `EtcFstab::fstab_encode()`.

aschnell:
So the security team answered and the hard coded path is not a security issue. Also a failure to write the log should not be a security issue here since the log is only for debugging.

BTW: YaST does also not check failures to write logs (https://github.com/yast/yast-core/blob/master/liby2util-r/src/y2log.cc#L345) and that is not only used for testsuites. I would also say that a full disk should not prevent users from using YaST.

joseivanlopez:
@shundhammer After your review, IRC discussions, etc, I have changed several things, so I have re-created the commits again.

storage/Filesystems/BtrfsSubvolumeImpl.cc
```diff
@@ -193,14 +193,14 @@ namespace storage
     MountOpts
     BtrfsSubvolume::Impl::get_default_mount_options() const
     {
-	return MountOpts(vector<string>({ "subvol=/" + path }));
+	return MountOpts(vector<string>({ "subvol=" + EtcFstab::fstab_encode("/" + path) }));
```
aschnell: The mount options are also used directly for the mount command and there the fstab encoding is wrong.

storage/Utils/ColumnConfigFile.h
```diff
@@ -118,6 +118,9 @@ class ColumnConfigFile : public CommentedConfigFile
 
 
     ColumnConfigFile();
+
+    ColumnConfigFile(int permissions);
```
shundhammer: We don't need another constructor for that. Simply provide a default value (say, 0644) for the _permissions_ argument, then we also have a parameter-less default constructor as a byproduct.

That avoids the duplication that we now have.

storage/Utils/CommentedConfigFile.cc
```diff
@@ -126,7 +130,7 @@ bool CommentedConfigFile::read( const string & filename )
 
     // reading is done via AsciiFile to the mockup playback and recording
 
-    AsciiFile ascii_file(filename);
+    AsciiFile ascii_file(filename, false, permissions);
```
shundhammer: Please comment what _false_ means here. I typically do it like this:

```C++
AsciiFile ascii_file(filename,
                     false,   // don't summon any monsters from their hiding place
                     permissions);
```

shundhammer:
Just the two minor things, otherwise LGTM.

shundhammer:
LGTM now once Travis is happy.

storage/EtcCrypttab.h
```diff
@@ -199,6 +199,10 @@ namespace storage
          * Dump the current contents to the log.
          **/
         void log();
+
+    private:
+
+	static const int DEFAULT_PERMISSIONS;
```
aschnell: The value can be set right here, see https://en.cppreference.com/w/cpp/language/static.

storage/Utils/Logger.cc
```diff
@@ -77,30 +72,21 @@ namespace storage
...
+	permissions(permissions)
     {
-    public:
-
-	virtual void write(LogLevel log_level, const std::string& component, const std::string& file,
-			   int line, const std::string& function, const std::string& content) override;
-
-    };
-
+	create_file_if_not_exists(filename, permissions);
```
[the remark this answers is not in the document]
joseivanlopez: So, we would call open each time something is written. Why is it better?

storage/Utils/LockImpl.cc
```diff
@@ -66,8 +66,8 @@ namespace storage
 		y2err("creating directory for lock-file failed: " << strerror(errno));
 	    }
 
-	    fd = open(LOCK_DIR "/lock", (read_only ? O_RDONLY : O_WRONLY) | O_CREAT | O_CLOEXEC,
-		      S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
+	    fd = open(LOCK_DIR "/lock", (read_only ? O_RDONLY : O_WRONLY) | O_CREAT | O_CLOEXEC, 0644);
```
aschnell: AFAIS this does not address the DoS issue, 0600 would be needed for that.

joseivanlopez: Umm, so the lock cannot be read by other user? Does it block the lock access?

storage/Utils/Logger.cc
```diff
@@ -113,9 +132,9 @@ namespace storage
 
 
     Logger*
-    get_logfile_logger()
+    get_logfile_logger(const std::string& filename)
     {
-	static LogfileLogger logfile_logger;
+	static LogfileLogger logfile_logger(filename);
```
aschnell: AFAIS this will only respect the filename the first time the function is called. Keeping a list of LogfileLogger would be possible but would be over-engineering from my point of view. Documenting the behaviour should be enough.

storage/Utils/Logger.cc
```diff
@@ -89,13 +90,31 @@ namespace storage
...
 
 
+    const int LogfileLogger::DEFAULT_PERMISSIONS = 0640;
+
+
+    LogfileLogger::LogfileLogger(const string& filename, int permissions) :
+	filename(filename),
+	permissions(permissions)
+    {
+	create_file_if_not_existing(filename, permissions);
```
aschnell: The audit suggests to use an open()/fdopen() sequence.

aschnell: The create_file_if_not_existing function could be removed. And fopen does likely nothing else than open/fdopen anyway.

Why the security team suggests it I cannot say.

joseivanlopez:
Some potential security problems have been detected with current implementation of `Storage::create_file_if_not_existing`. The sequence to create a file is: open the file with some permissions + close the file + use the file. An attacker could hijack the file once it is closed and change the permissions.

TODO: Discuss with Johannes if this is a problem in our case.

jreidinger:
@joseivanlopez How can attacker hijack that file when it has permission that do write only for root? If he can do it, it means he has root permission or there is serious security issue with file permissions.

storage/EtcCrypttab.h
```diff
@@ -199,6 +199,10 @@ namespace storage
          * Dump the current contents to the log.
          **/
         void log();
+
+    private:
+
+	static const int DEFAULT_PERMISSIONS = 0640;
```
jreidinger: btw can you comment in code why 0640 and not 0600?

storage/Utils/AsciiFile.cc
```diff
@@ -114,16 +120,23 @@ namespace storage
 	{
 	    y2mil("saving file " << name);
 
-	    ofstream file(name);
-	    classic(file);
+	    FILE* file = fdopen(open(name.c_str(), O_WRONLY | O_TRUNC | O_CREAT, permissions), "w");
```
aschnell: Checking return value of open is missing.

storage/Utils/AsciiFile.cc
```diff
@@ -114,16 +120,23 @@ namespace storage
...
-	    file.close();
+		fputs(line_with_break.c_str(), file);
+
+		if (ferror(file))
+		    ST_THROW(IOException(sformat("Saving file %s failed: %s", name, strerror(errno))));
+	    }
 
-	    if (!file.good())
-		ST_THROW(IOException(sformat("Saving file %s failed.", name)));
+	    if(fclose(file) == EOF)
```
aschnell: I suggest to check for ```!= 0```.

storage/Utils/AsciiFile.cc
```diff
@@ -114,16 +120,23 @@ namespace storage
...
+		y2war("closing file " << name << "failed");
```
aschnell: No, failure to write a file like this must be reported to the user. See bsc #1115801.

storage/Utils/AsciiFile.cc
```diff
@@ -114,16 +120,23 @@ namespace storage
...
 
 	    for (const string& line : lines)
-		file << line << '\n';
+	    {
+		string line_with_break = line + "\n";
 
-	    file.close();
+		fputs(line_with_break.c_str(), file);
+
+		if (ferror(file))
```
aschnell: I think the check can go outside the loop.

storage/Utils/Logger.cc
```diff
@@ -89,18 +95,34 @@ namespace storage
...
+    {
+    }
+
+
     void
     LogfileLogger::write(LogLevel log_level, const std::string& component, const std::string& file,
 			 int line, const std::string& function, const std::string& content)
     {
-	FILE* f = fopen("/var/log/libstorage-ng.log", "ae");
+	FILE* f = fdopen(open(filename.c_str(), O_WRONLY | O_APPEND | O_CREAT, permissions), "a");
```
aschnell: The "e" flag is missing.

storage/Utils/AsciiFile.cc
aschnell: I suggest to add FD_CLOEXEC and "e" here.

storage/Utils/Logger.cc
joseivanlopez: According to the documentation, "e" flag is ignored by fdopen.

storage/Utils/AsciiFile.cc
jreidinger: can we report here also why it failed? otherwise this log line has only limited usability.

storage/Utils/Logger.cc
aschnell: Thus the open call needs the flag. I would still add it to fdopen (for the developer).

storage/Utils/AsciiFile.h
```diff
@@ -55,10 +59,15 @@ namespace storage
 	vector<string>& get_lines() { return lines; }
 	const vector<string>& get_lines() const { return lines; }
 
+	void set_lines(const vector<string>& lines) { this->lines = lines; }
+
     protected:
 
+	static const int DEFAULT_PERMISSIONS = 0666;
```
jreidinger: please document why this so open permissions are ok as it looks very open.

storage/Utils/ColumnConfigFile.h
```diff
@@ -187,6 +187,10 @@ class ColumnConfigFile : public CommentedConfigFile
     vector<int> max_column_widths;
     int         max_column_width;
     bool        pad_columns;
+
+private:
+
+    static const int DEFAULT_PERMISSIONS = 0666;
```
jreidinger: same here.

storage/Utils/CommentedConfigFile.cc
```diff
@@ -138,30 +138,25 @@ void CommentedConfigFile::write( const string & new_filename )
...
+	this->filename = name;
 
     if ( name.empty() ) // Support for mocking:
-        return;         // Pretend everything worked just fine.
-
-    std::ofstream file( name, std::ofstream::out | std::ofstream::trunc );
-
-    if ( ! file.is_open() )
-	ST_THROW(IOException(sformat("Saving file %s failed.", name)));
+	return;		// Pretend everything worked just fine.
```
jreidinger: indentation looks a bit broken.

storage/Utils/AsciiFile.h
joseivanlopez: It is the default permissions for any file. The resulting file will have 666 - umask. With 666 we are simply using the default value that the system uses.

storage/Utils/CommentedConfigFile.h
```diff
@@ -520,9 +520,12 @@ class CommentedConfigFile: private boost::noncopyable
 
 private:
 
+    static const int DEFAULT_PERMISSIONS = 0666;
```
jreidinger: same here

storage/Utils/AsciiFile.cc
joseivanlopez: I have replaced it by an exception, as suggested by @aschnell.

storage/Utils/CommentedConfigFile.cc
joseivanlopez: I see it correctly ?

storage/EtcCrypttab.h
joseivanlopez: You are right, it should be 0600.

joseivanlopez:
@jreidinger To be honest, I don't see it a problem either, due to the new file would have 0644 at most. Anyway, now the file is created and written, and it is closed only at the end. So no way to any attack.

storage/Utils/Logger.cc
```diff
@@ -89,18 +95,35 @@ namespace storage
...
+    {
+    }
+
+
     void
     LogfileLogger::write(LogLevel log_level, const std::string& component, const std::string& file,
 			 int line, const std::string& function, const std::string& content)
     {
-	FILE* f = fopen("/var/log/libstorage-ng.log", "ae");
+	FILE* f = fdopen(open(filename.c_str(), O_WRONLY | O_APPEND | O_CREAT | O_CLOEXEC, permissions), "ae");
```
aschnell: Here the open return value should also be checked.

storage/Utils/Logger.h
```diff
@@ -82,10 +82,13 @@ namespace storage
 
     /**
      * Returns a Logger that logs to the standard libstorage log file
-     * ("/var/log/libstorage.log"). Do not use this function for production
-     * code but only for examples and test-cases.
+     * ("/var/log/libstorage.log") or to a given file. Do not use this
+     * function for production code but only for examples and test-cases.
+     *
+     * Note that this method only uses the given filename the fist time that
```
aschnell: typo: first

aschnell:
LGTM

PR #599, opened 2018-12-05
- use exceptions to handle errors

joseivanlopez:
Backport of #594


jreidinger:
as it is backwards incompatible change, I expect to see some conflicts in spec file that ensure that old yast2-storage-ng is not used with this modified library.

joseivanlopez:
@jreidinger I think change is backward compatible with current version of yast2-storage-ng (4.0.216) in SLE-15-GA.

We don't have bindings for the modified classes, and the new exception is correctly managed by storage-ng via callbacks, see https://github.com/yast/yast-storage-ng/blob/SLE-15-GA/src/lib/y2storage/storage_manager.rb#L310

PR #603, opened 2019-01-16
Add support for flash-only bcache

joseivanlopez:
PBI: https://trello.com/c/T7taTEix/606-5-use-the-bcache-command-for-probing

Note for reviewers: actions to create/delete Flash-only Bcache is out of scope. That will come with another PBI.


aschnell:
Since the change breaks the ABI a new major library version is required.

doc/bcache.md
```diff
@@ -0,0 +1,79 @@
...
+A Bcache device represents a block device that is using a cset for caching read/write operations. The
+Bcache device is created over a backing device, and the backing device can be any block device.
+
+A Bcache device could use none cset at all, so IO is directly performed over the backing device.
+
+A Bcache device supports all operations that can be performed over a block device, so it could be
+partitioned, encrypted, formatted, etc. A Backed Bcache device is represented by `BackedBcache` class
+in libstorage-ng.
+
+## Flash-only Bcache device
```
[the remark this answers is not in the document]
joseivanlopez: Nop AFAIK, they are completely different devices.

doc/autodocs/doxygen.conf
```diff
@@ -13,10 +13,12 @@ TAB_SIZE = 8
 
 INPUT = ../../storage
 FILE_PATTERNS = Actiongraph.h Devicegraph.h Environment.h FreeInfo.h Graphviz.h	\
-	SimpleEtcFstab.h SimpleEtcCrypttab.h Storage.h UsedFeatures.h Version.h
+	SimpleEtcFstab.h SimpleEtcCrypttab.h Storage.h UsedFeatures.h Version.h	\
+	CompoundAction.h
```
aschnell: CompoundAction.h does not include any documentation.

doc/bcache.md
```diff
@@ -0,0 +1,79 @@
+# Bcache
+
+Bcache technology allows to speed up the access time (read and write) for old hard disks. The idea
```
aschnell: "speed up the access time" sounds strange. I suggest to either use ""speed up the access" or "reduce the access time".

doc/autodocs/doxygen.conf
joseivanlopez: There is a new static method with `@throw DeviceNotFound` doc, so we need to include it in doxygen to generate the catches file for bindings.

storage/Devices/FlashBcacheImpl.cc
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact Novell, Inc.
+ *
+ * To contact Novell about this file by physical or electronic mail, you may
+ * find current contact information at www.novell.com.
+ */
+
+
+#include <regex>
```
aschnell: Seems unused.

LIBVERSION
```diff
@@ -1 +1 @@
-1.6.1
+2.0.0
```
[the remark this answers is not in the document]
joseivanlopez: I don't know what to touch in libstorage-ng.spec.in. In that file there is `Version:        @VERSION@` which is the package version, but not the library version.

I guess libstorage-ng1-4.1.79-0.x86_64.rpm is correct name for the package. In my TW I see libstorage-ng1-4.1.75-1.1.x86_64.

@wfeldt Could you please indicate what is exactly needed to touch to increase the major version of the library? Asking you because I see that libstorage-ng is using linuxrc tools for versioning.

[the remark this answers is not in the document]
aschnell: Change ```%define libname %{name}1``` to  ```%define libname %{name}2```. Without that building locally fails with rpmlint complaining about a shared library policy violation:

```
libstorage-ng1.x86_64: E: shlib-policy-name-error (Badness: 10000) libstorage-ng2
Your package contains a single shared library but is not named after its
SONAME.
```

storage/Devices/FlashBcacheImpl.cc
joseivanlopez: Removed. Thanks.

LIBVERSION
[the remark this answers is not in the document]
joseivanlopez: Thanks ;)

storage/CompoundAction/Formatter/Bcache.cc
```diff
@@ -400,4 +411,13 @@ namespace storage
                         mount_point.c_str() );
     }
 
+
+    const BlkDevice*
+    CompoundAction::Formatter::Bcache::get_blk_device() const {
+	if(is_backed_bcache(bcache))
+	    return static_cast<const storage::BackedBcache*>(bcache)->get_backing_device();
```
aschnell: Why not use the to_backed_bcache() function?

storage/CompoundAction/Formatter/Bcache.cc
```diff
@@ -400,4 +411,13 @@ namespace storage
...
+	else
+	    return static_cast<const storage::FlashBcache*>(bcache)->get_bcache_cset()->get_blk_devices().front();
```
aschnell: Simply calling front() without checking empty() looks dangerous to me.

aschnell:
LGTM

From my gut feeling I would have taken another approach: Not adding BackedBcache and FlashBcache but instead introduce a BcacheType for Bcache. That way it is done in LvmLv. That would likely also avoid the major API/ABI jump.

Maybe something for the PO to decide.

ancorgs:
Well, consistency is probably the most important feature of an API. If the WhateverTpe approach is used in the rest of the API, I would stick to it.

joseivanlopez:
Closing in favor of https://github.com/openSUSE/libstorage-ng/pull/607

PR #607, opened 2019-01-29
Add support for flash-only bcache (with ABI compatibility)

joseivanlopez:
PBI: https://trello.com/c/T7taTEix/606-5-use-the-bcache-command-for-probing

This PR adds support for probing so-called "Flash-only" Bcache devices. Initially, another PR #603 was proposed. Although API offered by #603 might be more precise (IMHO), it breaks the ABI compatibility due to the Bcache API is modified.

Due to we need to preserve ABI compatibility during the whole SLE15 cycle, another approach has been implemented here, where the ABI compatibility is kept.

PR #603 should be taken into account for the future, when ABI changes will be allowed (i.e., for SLE16).

Note for reviewers: actions to create/delete Flash-only Bcache devices is out of scope. That will come with another PBI.


storage/Devices/Bcache.cc
```diff
@@ -21,7 +21,8 @@
 
 
 #include "storage/Devices/BcacheImpl.h"
-#include "storage/Devicegraph.h"
+#include "storage/Devices/DeviceImpl.h"
+#include "storage/Utils/Enum.h"
```
aschnell: I do not see why this include should be needed.

storage/Devices/Bcache.h
```diff
@@ -33,80 +36,122 @@ namespace storage
     class BcacheCset;
 
 
+    /**
+     * Bcache types
+     */
+    enum class BcacheType {
+	BACKED, FLASHONLY
```
aschnell: I think FLASH_ONLY is better readable and also the standard style in C/C++.

storage/Devices/Bcache.h
```diff
@@ -33,80 +36,122 @@ namespace storage
...
+	 * This method does not make sense for Flash-only Bcache devices.
 	 *
 	 * @throw Exception
 	 */
 	void attach_bcache_cset(BcacheCset* bcache_cset);
 
 	/**
 	 * Returns cache mode attribute.
+	 *
+	 * This method does not make sense for Flash-only Bcache devices.
```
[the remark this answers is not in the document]
joseivanlopez: Ok. And what about the rest of attributes of backed Bcache? `sequential_cutoff` and `writeback_percent` do not make sense for Flash-only either.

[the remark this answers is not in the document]
joseivanlopez: BTW, in `LvmLv` I see that `set_chunk_size` only applies to thin pools, but methods `get_chunk_size` and `set_chunk_size` do not throw an exception when they are called over other types of Lvs.

Maybe in `Bcache` we should do the same for uniformity?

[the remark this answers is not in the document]
aschnell: OK, then for consistency leave it that way.

storage/Devices/Bcache.h
joseivanlopez: Done.

storage/Devices/Bcache.h
```diff
@@ -33,80 +36,122 @@ namespace storage
...
+	 * @throw Exception
 	 */
-	const BlkDevice* get_blk_device() const;
+	const BlkDevice* get_backing_device() const;
 
 	/**
-	 * Returns true if a BcacheCset is attached.
+	 * @copydoc get_backing_device
+	 */
+	const BlkDevice* get_blk_device() const ST_DEPRECATED { return get_backing_device(); }
```
aschnell: I discussed this with Stefan and we are not sure whether implementing the function in the header does not break the ABI. E.g. the compiler could decide to not include the function in the library since it is inlined.

joseivanlopez: Ok, makes sense. Moved to .cc file.

aschnell:
LGTM

PR #610, opened 2019-02-01
- check for some storage signatures next to partition tables

aschnell:
For https://trello.com/c/qLpplZWL/656-1-sles15-sp1-p2-1115807-disk-as-pv-is-not-added-to-generate-ay-profile-when-cloning-first-approach, https://bugzilla.suse.com/show_bug.cgi?id=1115807.


aschnell:
The po files are generated by 'make merge' in the po directory.

PR #611, opened 2019-02-01
Use bcache command

joseivanlopez:
## Problem

Actions for Bcache devices are performed via `echo 1 > /sys/..`, but `bcache-tools` package now offers a new command to work with Bcache devices. libstorage-ng should use such command instead of directly touch files at `/sys`.

The library should only use `bcache` command to perform the actions, but such command has no way to make permanent changes for some attributes (`writeback_percent` and `sequential_cutoff`). Even more, when those values are "manually" set with "echo" to the proper "/sys" file, the values are also reset after each restart.

* https://trello.com/c/sdR0Hwi3/607-2-use-the-new-bcache-command-instead-of-make-bcache
* https://fate.suse.com/325346

## Solution

Replace `echo` commands by `bcache` equivalents. Moreover, this PR removes setters for non-permanent stored values from the public API. Note that, for time being, both getters and setters are still available in private classes. Most likely, in future, `bcache` command will allow to permanent save such attributes.

## Testing

Tested manually.


storage/Devices/BcacheCsetImpl.cc
```diff
@@ -336,11 +336,17 @@ namespace storage
     void
     BcacheCset::Impl::do_deactivate() const
     {
-	string cmd_line = ECHO_BIN " 1 > " + quote(SYSFS_DIR "/fs/bcache/" + get_uuid() + "/stop");
+	// TODO handle multiple BlkDevices?
+
+	const BlkDevice* blk_device = get_blk_devices().front();
```
shundhammer: What if this is empty? I.e. if there is no block device at all?
Even if you are sure that this should not happen, the code should not crash at this point.

storage/Devices/BlkDeviceImpl.cc
```diff
@@ -825,4 +867,46 @@ namespace storage
...
+    {
+	SystemCmd(UDEVADMBIN_SETTLE);
+
+	for (auto name : dev_names)
+	{
+	    bool exists = access(name.c_str(), R_OK) == 0;
+	    y2mil("name:" << name << " exists:" << exists);
+
+	    if (exists)
+	    {
```
shundhammer: Please add an info/milestone logging here so we know what's going on when we get bug report that YaST hangs. Since this might wait for quite a while, it will be unclear what's going on here, and it will be very hard to debug without such a logging line.

storage/Devices/BcacheImpl.cc
```diff
@@ -615,29 +615,19 @@ namespace storage
 
 	if(get_type() == BcacheType::BACKED)
 	{
-	    const BlkDevice* blk_device = get_backing_device();
+	    const BlkDevice* backing_device = get_backing_device();
```
shundhammer: Please add a null check here to avoid a crash if for some reason there is no backing device.

storage/Devices/BlkDeviceImpl.cc
```diff
@@ -825,4 +867,46 @@ namespace storage
 	}
     }
 
+
+    void
+    wait_for_detach_devices(const vector<const BlkDevice*>& blk_devices)
+    {
+	vector<string> dev_names;
+
+	for_each(blk_devices.begin(), blk_devices.end(), [&dev_names](const BlkDevice* dev)
```
shundhammer: Why not use a simple `for` loop here? You are iterating over the entire container from begin() to end().
I might be overlooking something here, but this seems awfully complicated for just collecting those names.

storage/Devices/BlkDeviceImpl.h
```diff
@@ -211,6 +221,13 @@ namespace storage
      */
     void wait_for_devices(const vector<const BlkDevice*>& blk_devices);
 
+
+    /**
+     * Run "udevadm settle" and check non existence of all blk devices.
+     */
+    void wait_for_detach_devices(const vector<const BlkDevice*>& blk_devices);
```
shundhammer: Should this really be an official part of the API? This sounds very much like an implementation detail to me.

storage/Devices/BcacheImpl.cc
joseivanlopez: There is already a check for that, `get_backing_device` method throws an `DeviceNotFound` exception when there is no backing device.

storage/Devices/BlkDeviceImpl.h
joseivanlopez: Well, this is declared in a Impl header, so I expect this does not belong to the public API. Anyway, basically I have mimic the method `wait_for_devices` that is already there. This new `wait_for_detach_devices` is just the opposite method, it waits to devices disappear instead of waiting for devices being created.

storage/Devices/Bcache.h
```diff
@@ -132,31 +132,13 @@ namespace storage
...
 	unsigned long long get_sequential_cutoff() const;
 
-	/**
-	 * Sets sequential_cutoff attribute.
-	 *
-	 * This method does not make sense for Flash-only Bcache devices.
-	 *
-	 * @param[in] size size in a bytes
-	 */
-	void set_sequential_cutoff(unsigned long long size);
```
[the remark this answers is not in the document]
joseivanlopez: Well, those methods were added after SLE-15-GA release, so nobody should be using them. Only TW users could be using it, but we agreed to ignore TW this time due to is quite unlikely that somebody is using it.

So, as summary, we introduced some methods during SP1 development but then we have removed them before SP1 release, so nothing has happened here :).

storage/Devices/BlkDeviceImpl.h
shundhammer: Ah, duh, silly me; I thought I had checked that this is the public header. Disregard.

storage/Devices/Bcache.h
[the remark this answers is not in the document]
shundhammer: OK then.

storage/Devices/BcacheImpl.cc
shundhammer: OK.

storage/Devices/BcacheImpl.cc
```diff
@@ -725,14 +715,9 @@ namespace storage
     void
     Bcache::Impl::do_attach_bcache_cset() const
     {
-	const BcacheCset* bcache_cset = get_bcache_cset();
-
-	string cmd_line = ECHO_BIN " " + quote(bcache_cset->get_uuid()) + " > " +
-	    quote(SYSFS_DIR "/" + get_sysfs_path() + "/bcache/attach");
+	string cmd_line = BCACHE_BIN " attach " + quote(get_bcache_cset()->get_uuid()) + " " + quote(get_backing_device()->get_name());
```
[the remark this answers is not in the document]
joseivanlopez: Answered above.

shundhammer:
LGTM now.

There is now just that one wait...() logging left; if this now happens for all kinds of devices all the time, I'd remove that first logging again (but keep the one that is done after some repeated timeouts). If it's only for the Bcache case, simply keep it.

PR #613, opened 2019-02-12
Allow to modify Bcache devices

joseivanlopez:
PBI: https://trello.com/c/cQo16PQ5/672-editing-an-in-memory-bcache-device-devices-and-cache-mode


storage/Devices/Bcache.cc
```diff
@@ -125,9 +125,16 @@ namespace storage
 
 
     void
-    Bcache::attach_bcache_cset(BcacheCset* bcache_cset)
+    Bcache::add_bcache_cset(BcacheCset* bcache_cset)
```
aschnell: What is the motivation for this change? The term "attach" is used at may places in this concern, e.g. by the bcache tool (as seen in Bcache::Impl::do_attach_bcache_cset()).

storage/Devices/BcacheImpl.cc
```diff
@@ -407,6 +407,21 @@ namespace storage
     }
 
 
+    void
+    Bcache::Impl::remove_bcache_cset()
+    {
+	if(get_type() == BcacheType::FLASH_ONLY)
+	    ST_THROW(LogicException("A Caching Set cannot be detached from a Flash-only Bcache"));
```
aschnell: Here the function is called "remove_..." but the error message says "detach". Please be consistent with the terms (I suggest to follow the bcache project).

storage/Devices/Bcache.cc
joseivanlopez: As discussed on IRC, this was changed for consistency with the rest of the public API.

storage/Devices/BcacheImpl.cc
joseivanlopez: I will adapt the messages.

storage/Devices/Bcache.h
```diff
@@ -101,13 +101,22 @@ namespace storage
...
 	/**
-	 * Attach a caching set to the Bcache.
+	 * Associates a caching set to the Bcache.
 	 *
 	 * This method does not make sense for Flash-only Bcache devices.
 	 *
 	 * @throw LogicException, Exception
 	 */
-	void attach_bcache_cset(BcacheCset* bcache_cset);
+	void add_bcache_cset(BcacheCset* bcache_cset);
```
aschnell: I assume the ABI break here is considered to be OK.

storage/Devices/BcacheImpl.cc
```diff
@@ -576,6 +591,42 @@ namespace storage
     }
 
 
+    void
+    Bcache::Impl::add_modify_actions(Actiongraph::Impl& actiongraph, const Device* lhs) const
+    {
+	const Bcache* lhs_bcache = to_bcache(lhs);
+
+	// TODO Flash-only Bache
```
aschnell: typo "bache"

storage/Devices/BcacheImpl.cc
```diff
@@ -725,6 +776,117 @@ namespace storage
...
+	return sformat(text, toString(get_cache_mode()), get_name(), get_size_text());
+    }
+
+
+    void
+    Bcache::Impl::do_update_cache_mode() const
+    {
+	string cmd_line = BCACHE_BIN " set-cachemode " + quote(get_backing_device()->get_name()) + " " + quote(toString(get_cache_mode()));
+
+	SystemCmd cmd2(cmd_line, SystemCmd::DoThrow);
```
aschnell: The "2" in "cmd2" is confusing.

aschnell:
Only some minor stuff, LGTM.

storage/Devices/Bcache.h
joseivanlopez: Yes, because `attach_bcache_cset` was added during SP1 development phase. So we are still in time to break it.

PR #615, opened 2019-02-19
Workaround upgrade with multiple mout points per block device (bsc#11…

jreidinger:
…18865)


storage/EtcFstab.cc
```diff
@@ -690,16 +690,29 @@ namespace storage
...
+	    // for multiple mount points for same partition try at first identical path
+	    // and if not found, then any
 	    vector<JointEntry>::iterator it = find_if(ret.begin(), ret.end(), [&path](const JointEntry& tmp) {
                 bool in_fstab = tmp.is_in_etc_fstab();
                 if (in_fstab && tmp.fstab_entry->get_mount_point() != path)
                 {
-                    y2war("mount points for " << tmp.fstab_entry->get_device()
-                          << " differ: fstab(" << tmp.fstab_entry->get_mount_point()
-                          << ") != proc(" << path << ")");
+		    return false;
```
joseivanlopez: NP: It was a little bit difficult to understand this to me. Why not a more explicit comparison?

```
if (!in_fstab)
  return false;

tmp.fstab_entry->get_mount_point() == path
```

storage/Filesystems/BlkFilesystemImpl.cc
```diff
@@ -267,7 +268,15 @@ namespace storage
 
 	vector<JointEntry> joint_entries = join_entries(fstab_entries, mount_entries);
 	if (!joint_entries.empty())
-	    joint_entries[0].add_to(get_non_impl());
+	{
+	    y2war("fstab entries: " << fstab_entries);
+	    y2war("mount entries: " << mount_entries);
+	    if (joint_entries.size() > 1)
+		y2war("More then one entry for given aliases: " << aliases);
```
joseivanlopez: Typo: "More than"

PR #621, opened 2019-03-04
handle entries in crypttab whose block device do not exist

aschnell:
For https://trello.com/c/69RMJz4r/685-sles15-sp1-p3-1123342-storage-ng-error-when-probing-a-non-existent-luks-device or https://bugzilla.suse.com/show_bug.cgi?id=1123342.


storage/EtcCrypttab.cc
```diff
@@ -246,8 +246,18 @@ namespace storage
...
 	    if (boost::starts_with(blk_device, DEV_DIR "/"))
 	    {
-		if (system_info.getCmdUdevadmInfo(blk_device).get_majorminor() == majorminor)
-		    return entry;
+		try
+		{
+		    if (system_info.getCmdUdevadmInfo(blk_device).get_majorminor() == majorminor)
+			return entry;
+		}
+		catch (const Exception& e)
```
lslezak: Just a question: Would it be possible to catch just some specific exceptions? I mean this could potentially hide some real errors.

aschnell: The command udevadm may not exist. That could be handled by first catching CommandNotFoundException and rethrowing that exception. But since udevadm is use at so many places no probing will work without it so handling this here seems over-engineered to me.

lslezak: OK.

PR #628, opened 2019-03-12
improved calculation of usable size for LVM PVs (bsc#1121129)

aschnell:
For https://trello.com/c/37u3G3aN/ or https://bugzilla.suse.com/show_bug.cgi?id=1121129.

This adds topology handling to Partition and Luks, esp. optimal_io_size, which is needed to calculate the usable size of LVM PVs.


shundhammer:
LGTM

PR #629, opened 2019-03-13
Unifying name Bcache/bcache to bcache

schubi2:


storage/Devices/BcacheCsetImpl.cc
```diff
@@ -248,11 +248,11 @@ namespace storage
 			   // TRANSLATORS: displayed before action,
 			   // %1$s is replaced by device name (e.g. /dev/sda1),
 			   // %2$s is replaced by size (e.g. 2 GiB)
-			   _("Create Bcache cache set on %1$s (%2$s)"),
+			   _("Create Bcache Cache Set on %1$s (%2$s)"),
```
jreidinger: here it is still Bcache also bug cache set looks strange, not? or is it title?

storage/Devices/BcacheImpl.cc
```diff
@@ -555,7 +555,7 @@ namespace storage
 	    [&actiongraph, &devicegraph_rhs](Actiongraph::Impl::vertex_descriptor vertex) {
 	    const Action::Base* action = actiongraph[vertex];
 	    const Bcache* bcache = to_bcache(devicegraph_rhs->find_device(action->sid));
-	    return bcache->get_number();
+	    return cache->get_number();
```
jreidinger: this looks from as above it is bcache.

PR #630, opened 2019-03-15
- fix compilation with -O0

aschnell:
See https://stackoverflow.com/questions/3025997/defining-static-const-integer-members-in-class-definition.

### DIFF
--- a/storage/Devices/LvmPvImpl.cc
+++ b/storage/Devices/LvmPvImpl.cc
@@ -46,6 +46,8 @@ namespace storage
 
     const char* DeviceTraits<LvmPv>::classname = "LvmPv";
 
+    const unsigned long long LvmPv::Impl::default_pe_start;
+
 
     LvmPv::Impl::Impl(const xmlNode* node)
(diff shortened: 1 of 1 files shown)


storage/Devices/LvmPvImpl.cc
```diff
@@ -46,6 +46,8 @@ namespace storage
 
     const char* DeviceTraits<LvmPv>::classname = "LvmPv";
 
+    const unsigned long long LvmPv::Impl::default_pe_start;
```
jreidinger: shouldn't be const initialized here?

jreidinger: or better why it is needed also here, when it is already initialized in header? I still do not get why this change is needed.

aschnell: In C/C++ there is the declaration and the definition of objects. The header provides the declaration (maybe with initialization) but not the definition. Normally both declaration and definition are needed.

What is special here is that with optimization the compiler generates code that does not need the address of the variable (thus it does not need the definition). Since I normally compile with optimization I did not notice the missing definition right away.

PR #631, opened 2019-03-22
Use the libstorage-ng SLE15 (Leap 15.0) image in Travis

lslezak:
- Use the libstorage-ng specific Docker image
- Use the SLE15/Leap 15.0 based image (instead of the Tumbleweed one)


aschnell:
LGTM

PR #634, opened 2019-04-11
- work on btrfs with multiple devices

aschnell:
For https://trello.com/c/p88ZCmfO/920-3-libstorage-ng-probing-multi-device-btrfs.


storage/Filesystems/BlkFilesystemImpl.cc
```diff
@@ -708,43 +728,40 @@ namespace storage
...
-			   // %2$s is replaced by label (e.g. ROOT)
-			   _("Setting label of %1$s to %2$s"));
+			   // %1$s is replaced by file system name (e.g. ext4),
+			   // %2$s is replaced by one or more devices (e.g /dev/sda1 (1.0 GiB) and
+			   // /dev/sdb2 (1.0 GiB)),
+			   // %3$s is replaced by label (e.g. ROOT)
+			   _("Setting label of %1$s on %2$s to %3$s"));
 
-	return sformat(text, blk_device->get_name(), label);
+	return sformat(text, get_displayname(), join(get_blk_devices(), JoinMode::COMMA, 10), label);
```
shundhammer: Since this `join(...)` is used so many times, maybe add a function for it? 

```C++
string get_blk_devices_list( join_mode, int max_count );
```

?

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -88,6 +113,28 @@ namespace storage
     }
 
 
+    FilesystemUser*
+    Btrfs::Impl::add_device(BlkDevice* blk_device)
+    {
+	if (blk_device->num_children() != 0)
```
shundhammer: I guess this check will stay like this only during a transitional period, right?
In the future, it will need to check if the resulting number of children matches the BtrfsRaidLevel.
Like this, it only works with one single block device.

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -88,6 +113,28 @@ namespace storage
...
+	    ST_THROW(WrongNumberOfChildren(blk_device->num_children(), 0));
+
+	FilesystemUser* filesystem_user = FilesystemUser::create(get_devicegraph(), blk_device, get_non_impl());
+
+	return filesystem_user;
+    }
+
+
+    void
+    Btrfs::Impl::remove_device(BlkDevice* blk_device)
```
shundhammer: No check if `blk_device` is actually part of this Btrfs?
The subsequent instructions in the function may fail in spectacular ways if a caller passes a wrong block device or even a null pointer.

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -241,11 +293,96 @@ namespace storage
...
+    {
+	if (!support_btrfs_multiple_devices())
+	    return;
+
+	SystemInfo& system_info = prober.get_system_info();
+	Devicegraph* system = prober.get_system();
+
+	const CmdBtrfsFilesystemShow& cmd_btrfs_filesystem_show = system_info.getCmdBtrfsFilesystemShow();
+
+	for (const CmdBtrfsFilesystemShow::value_type& tmp : cmd_btrfs_filesystem_show)
```
shundhammer: `tmp` is not a good variable name for this; this is the loop variable for this entire longish loop. There must be a better name for it. `btrfs` comes to mind. Or maybe `detected_btrfs`. But this variable is really the pivotal point of this whole block, so it deserves some more attention, i.e. a speaking name.

shundhammer:
LGTM in general. My only real concern is that `tmp` (or `tmp.second`) variable name.

storage/Filesystems/BtrfsImpl.cc
aschnell: This check ensures that the block device being added has no child yet, so is unused.

storage/Filesystems/BtrfsImpl.cc
aschnell: The find_holder will throw an exception if the block device is not part of the btrfs.

I will add checks for null.

storage/Filesystems/BlkFilesystemImpl.cc
aschnell: That does not look really shorter to me.

PR #637, opened 2019-04-17
- extended documentation

aschnell:


storage/Filesystems/Btrfs.h
```diff
@@ -94,15 +94,17 @@ namespace storage
 	 * Get the allowed metadata RAID levels for the btrfs. So far,
 	 * this depends on the number of devices. Levels for which
 	 * mkfs.btrfs warns that they are not recommended are not
-	 * considered allowed here.
+	 * considered allowed here. Additionally DEFAULT is allowed
```
lslezak: Outside of your change, but this wording looks strange `... they are not recommended are not considered allowed here`, I guess just one of considered/allowed should be here.

lslezak: The same below.

PR #638, opened 2019-04-18
- parse devid in output of 'btrfs filesystem show' 

aschnell:
Also parse the devid in the output of 'btrfs filesystem show'. The devid will be required for some btrfs commands when adding multiple devices support.

### DIFF
--- a/storage/SystemInfo/CmdBtrfs.cc
+++ b/storage/SystemInfo/CmdBtrfs.cc
@@ -69,6 +69,10 @@ namespace storage
     void
     CmdBtrfsFilesystemShow::parse(const vector<string>& lines)
     {
+	static const regex uuid_regex("uuid: (" UUID_REGEX ")", regex::extended);
+
+	smatch match;
+
 	vector<string>::const_iterator it = lines.begin();
 
 	while (it != lines.end())
@@ -81,7 +85,11 @@ namespace storage
 		y2mil( "uuid line:" << *it );
 
 		Entry entry;
-		entry.uuid = extractNthWord( 3, *it );
+
+		if (!regex_search(*it, match, uuid_regex))
+		    ST_THROW(Exception("did not find uuid"));
+
+		entry.uuid = match[1];
 		y2mil("uuid:" << entry.uuid);
 
 		++it;
@@ -92,9 +100,15 @@ namespace storage
 		while( it!=lines.end() && boost::contains( *it, "devid " ) )
 		{
 		    y2mil( "devs line:" << *it );
-		    string device = extractNthWord( 7, *it );
-		    if ( !boost::contains( device, "/dev" ) )  // Allow /sys/dev or /proc/devices
-			ST_THROW( ParseException( "Not a valid device name", device, "/dev/..." ) );
+
+		    Device device;
+
+		    extractNthWord(1, *it) >> device.id;
+
+		    device.name = extractNthWord(7, *it);
+		    if (!boost::contains(device.name, DEV_DIR "/"))  // Allow /sys/dev or /proc/devices
+			ST_THROW( ParseException( "Not a valid device name", device.name, "/dev/..." ) );
+
 		    entry.devices.push_back( device );
 		    ++it;
 		}
@@ -106,6 +120,7 @@ namespace storage
 		}
 
 		y2mil("devices:" << entry.devices);
+
 		data.push_back(entry);
 	    }
 	}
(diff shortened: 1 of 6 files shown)


storage/SystemInfo/CmdBtrfs.cc
```diff
@@ -92,9 +100,15 @@ namespace storage
...
 		    y2mil( "devs line:" << *it );
-		    string device = extractNthWord( 7, *it );
-		    if ( !boost::contains( device, "/dev" ) )  // Allow /sys/dev or /proc/devices
-			ST_THROW( ParseException( "Not a valid device name", device, "/dev/..." ) );
+
+		    Device device;
+
+		    extractNthWord(1, *it) >> device.id;
+
+		    device.name = extractNthWord(7, *it);
```
jreidinger: when you touch it, can you please add example output as commend to the begging of this parse? It can really helps with all those word extraction to see what is each of that trying to parse.

aschnell: We have the testsuite with several examples.

storage/SystemInfo/CmdBtrfs.cc
```diff
@@ -81,7 +85,11 @@ namespace storage
 		y2mil( "uuid line:" << *it );
 
 		Entry entry;
-		entry.uuid = extractNthWord( 3, *it );
+
+		if (!regex_search(*it, match, uuid_regex))
```
jreidinger: if I get code this can only happen when line contain ` uuid: ` but not UUID_REGEX after it. So malformed output from btrfs?

storage/Utils/StorageDefines.h
```diff
@@ -133,4 +133,7 @@
 #define MKFS_UDF_BIN "/usr/sbin/mkfs.udf"
 
 
+#define UUID_REGEX "[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}"
```
jreidinger: just question: storage never use uppercase in uuid? As it is generic UUID regex it can be problem if it is used on uuid that contain uppercase letters.

testsuite/SystemInfo/btrfs-filesystem-show.cc
```diff
@@ -81,9 +81,9 @@ BOOST_AUTO_TEST_CASE(parse_good)
     };
 
     vector<string> output = {
-	"uuid:ea108250-d02c-41dd-b4d8-d4a707a5c649 devices:</dev/mapper/system-test>",
-	"uuid:d82229f2-f9e4-40fd-b15f-84e2d42e6d0d devices:</dev/mapper/system-testsuite>",
-	"uuid:653764e0-7ea2-4dbe-9fa1-866f3f7783c9 devices:</dev/mapper/system-btrfs>"
+	"uuid:ea108250-d02c-41dd-b4d8-d4a707a5c649 devices:<{ id:1 name:/dev/mapper/system-test }>",
+	"uuid:d82229f2-f9e4-40fd-b15f-84e2d42e6d0d devices:<{ id:1 name:/dev/mapper/system-testsuite }>",
+	"uuid:653764e0-7ea2-4dbe-9fa1-866f3f7783c9 devices:<{ id:1 name:/dev/mapper/system-btrfs }>"
```
jreidinger: previous output was wrong? If it was valid then maybe it helps to test also other format without id as now I see only outputs with id in it.

testsuite/SystemInfo/btrfs-filesystem-show.cc
```diff
@@ -149,7 +149,7 @@ BOOST_AUTO_TEST_CASE(systemcmd_error)
 BOOST_AUTO_TEST_CASE(parse_missing)
 {
     vector<string> input = {
-	"Label: none  uuid: b0749dbe-7de5-4719-9cb6-043dd5c70d00",
+	"Label: 'hello world'  uuid: b0749dbe-7de5-4719-9cb6-043dd5c70d00",
```
jreidinger: :+1: good to test this kind of words.

testsuite/SystemInfo/btrfs-filesystem-show.cc
jreidinger: ah, nevermind. I see that it is output of storage and not cmd output. cmd output is called input which confuse me, but it is consistent so fine for me.

storage/SystemInfo/CmdBtrfs.cc
aschnell: Yes, if the regex is not found it looks like a wrong output from btrfs (or the output has changed on purpose).

storage/Utils/StorageDefines.h
aschnell: I cannot say the storage never uses uppercase UUIDs but so far there is always lowercase.

testsuite/SystemInfo/btrfs-filesystem-show.cc
aschnell: This is the output after parsing to verify that the parsing work correctly. The "id" is added in the PR.

PR #640, opened 2019-04-23
- parse UUID in output of mkfs.btrfs

aschnell:
Parsing the UUID in the output of mkfs.btrfs avoids calling blkid.


wfeldt:
lgtm

PR #646, opened 2019-05-10
Resize multi-device Btrfs

joseivanlopez:
* Part of: https://jira.suse.de/browse/SLE-3877
* Part of PBI: https://trello.com/c/mB6C51ir/973-3-libstorage-ng-resize-btrfs-with-multiple-devices


storage/Filesystems/BtrfsImpl.cc
```diff
@@ -59,14 +60,16 @@ namespace storage
 
     Btrfs::Impl::Impl()
 	: BlkFilesystem::Impl(), configure_snapper(false), snapper_config(nullptr),
-	  metadata_raid_level(BtrfsRaidLevel::DEFAULT), data_raid_level(BtrfsRaidLevel::DEFAULT)
+	  metadata_raid_level(BtrfsRaidLevel::DEFAULT), data_raid_level(BtrfsRaidLevel::DEFAULT),
```
shundhammer: NP: Since this initializer list now spans more than one line anyway, it would make sense to put each one on a separate line so the second one doesn't get overlooked that easily.

Same for the other constructor.

shundhammer: OTOH you can save the effort of just using the default constructor; this is what happens anyway for non-primitive data types (anything else than char, int, bool) when you don't explicitly write it down.

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -505,18 +508,159 @@ namespace storage
...
+	{
+	    // The filesystem already exists on disk.
+
+	    const Btrfs* filesystem = redirect_to_system(get_non_impl());
+
+	    if (filesystem->get_blk_devices().size() == 1)
+	    {
+		// Currently, the filesystem is a single-device Btrfs.
+		
+		if (blk_device && blk_device->get_sid() != filesystem->get_impl().get_blk_device()->get_sid())
```
shundhammer: Hm... can we be sure that the blk_device that was passed actually belongs to this device?
I trust nobody, in particular not anybody calling my function... :smile:

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -505,18 +508,159 @@ namespace storage
...
+	    resize_info.min_size *= 1.5;
+
+	    resize_info.reasons |= RB_SHRINK_NOT_SUPPORTED_BY_MULTIDEVICE_FILESYSTEM;
+
+	    if (resize_info.min_size >= resize_info.max_size)
+		resize_info.reasons |= RB_FILESYSTEM_FULL;
+	}
+	else
+	{
+	    // Checking from an specific block device.
```
shundhammer: **a* specific block device

storage/Filesystems/BtrfsImpl.cc
```diff
@@ -505,18 +508,159 @@ namespace storage
...
+
+	    if (contain_device(get_blk_devices(), blk_device))
+	    {
+		// Block shrink only if the filesystem is already using this device on system.
```
shundhammer: on **the** system

shundhammer:
Besides the few minor things LGTM.

But I really can't say if the actual resize test / resize execution do the right thing, so we'll still need @aschnell to have another look at this.

storage/Filesystems/BtrfsImpl.cc
joseivanlopez: Yes, I know. I am simply doing the same as the rest of constructors, see for example `BlkFilesystem::Impl::Impl()`. We are always explicitly calling the constructor for each attribute, so I think we should continue doing the same (to not mix styles).

storage/Filesystems/BtrfsImpl.cc
joseivanlopez: Well, for now, this parameter only contains a parent of the filesystem. But yes, I would add an extra check.

joseivanlopez:
Last week I was talking with @aschnell about these changes and I have already done the suggested improvements (sentence when growing Btrfs and simply xml file for the tests). I proceed to merge them.

PR #648, opened 2019-05-23
- work on btrfs multiple device support

aschnell:
Preparations for https://trello.com/c/WQR8Xpou/1008-extend-fstab-btrfs-handling.


LIBVERSION
```diff
@@ -1 +1 @@
-1.10.3
+1.11.0
```
joseivanlopez: Is it necessary to increase the minor number? No changes in public API, isn't it?

aschnell: The two operator== in Remote.h are new.

PR #649, opened 2019-05-24
- fix resizing btrfs on e.g. LVM

aschnell:
The devid function does not handle device name aliasing, e.g. it cannot find /dev/test/foo which btrfs reports as /dev/mapper/test-foo. By using the newly added id from the devicegraph the function is not needed anymore.


wfeldt:
lgtm

PR #652, opened 2019-06-04
- added unit test for LVM over MD

aschnell:
Added a unit test to avoid (make less likely) regressions when changing anything for bsc #1136641.


shundhammer:
LGTM

PR #659, opened 2019-07-25
LUKS2 support

aschnell:
For https://trello.com/c/SGwB0orq/1155-libstorage-ng-support-for-luks2.


storage/Devices/Luks.h
```diff
@@ -52,8 +52,38 @@ namespace storage
...
+	 */
+	const std::string& get_format_options() const;
+
+	/**
+	 * Set extra options for luks format call.
+	 *
+	 * Options that modify the size of the resulting blk device
+	 * (e.g. --integrity) are not allowed.
+	 */
+	void set_format_options(const std::string& format_options);
```
lslezak: I do not know the general libstorage expectations but as the format options are literately injected into the command (no escaping/quoting) I think it should be mentioned in the comment that it's the responsibility of the caller to properly escape/quote the options.

aschnell:
Thanks for the suggested improvements.

PR #661, opened 2019-08-21
Fix surrounding detection

joseivanlopez:
Method `Partition#get_unused_surrounding_region` does not work properly when the next partition starts before 1 MiB (minimal MBR gap).

* https://trello.com/c/Pkbj2L2m/1230-tw-leap-151-storage-ng-crashes-when-trying-to-resize-a-partition


shundhammer:
Summarizing a discussion in IRC:

I was wondering if this shouldn't be moved to the _usable region_ or if this changes the definition of that usable region. Ivan wrote that it's slightly different on GPT vs. MS-DOS disk label. Ancor pointed to recent documentation here:

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/partition_tables/msdos.rb

shundhammer:
With the recent explanations LGTM.

shundhammer:
Even better now. Thanks for the explanation!

LGTM

PR #673, opened 2019-09-30
handle cipher and key size for encryption

aschnell:
For https://trello.com/c/B72IxUOM/1294-3-probing-of-cipher-and-recognize-preexisting-pervasive-encryption.


storage/SystemInfo/CmdCryptsetup.cc
```diff
@@ -119,15 +120,123 @@ namespace storage
...
+	}
+
 	y2mil(*this);
     }
 
 
+    void
+    CmdCryptsetupLuksDump::parse_version1(const vector<string>& lines)
+    {
+	static const regex cipher_name_regex("Cipher name:[ \t]*([^ \t]+)[ \t]*", regex::extended);
```
wfeldt: And also in all regexps below: the terminating "[ \t]*" part has no effect whatsoever.

wfeldt:
Basically, lgtm but I would simplify the regexps.

NP: why store the key size in bytes when it's used everywhere in bits?

storage/SystemInfo/CmdCryptsetup.cc
aschnell: No, regex_match tries to match the complete string. When using regex_search it would be indeed not needed.

aschnell:
Since the on-disk format of LUKS2 also stores the key size in bytes I have implemented it the same way now.

storage/SystemInfo/CmdCryptsetup.cc
wfeldt: Talked to Arvin; the regexps are correct as-is.

wfeldt:
lgtm

PR #675, opened 2019-10-18
Log messages about the mount_by fallback are just warnings

ancorgs:
Since QA recently complained about it, I took the opportunity to fix it as part of 

https://trello.com/c/GrmrcH1n/1337-3-sles15-sp1-p3-1151075-sles-15-sp1-yast-mounts-nvme-partitions-by-kernel-name-per-default-upon-reboot-they-are-randomly-enumera

I have not tested, but it looks trivial enough to be safe.

### DIFF
--- a/storage/Devices/BlkDeviceImpl.cc
+++ b/storage/Devices/BlkDeviceImpl.cc
@@ -277,25 +277,25 @@ namespace storage
 	switch (mount_by_type)
 	{
 	    case MountByType::UUID:
-		y2err("no uuid possible, using fallback");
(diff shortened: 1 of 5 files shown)


aschnell:
There are also messages in BlkFilesystemImpl.cc and SwapImpl.cc.

ancorgs:
Oops, I overlooked those two. I have fixed it via `git push -f`

aschnell:
LGTM

Without a bug/fate/jira number the various tools might reject it until other changes some.

PR #676, opened 2019-10-31
- handle is_permanent() in possible_mount_bys() (bsc#1155566)

aschnell:
See https://bugzilla.suse.com/show_bug.cgi?id=1155566. No PBI.


shundhammer:
LGTM

PR #677, opened 2019-11-12
Split MountPoint#set_default_mount_options into pieces (needed for jsc#SLE-7687)

ancorgs:
## Problem

In yast-storage-ng we want to automatically propose the `_netdev` option in fstab when needed (and to propagate that option to crypttab). See https://github.com/yast/yast-storage-ng/pull/995 for the details.

In its current implementation, libstorage-ng calculates a initial set of default options for fstab, but it does not provide a method to query them, just a method to directly set them: `MountPoint#set_default_mount_options`. That's problematic for the current implementation of yast-storage-ng.

- Current Jira entry: https://jira.suse.com/browse/SLE-7687
- Original bug report (that was then moved to another, and then to Fate, before ending up in the Jira mentioned above): https://bugzilla.suse.com/show_bug.cgi?id=1047099

## Solution

Split the existing method into two pieces for better granularity. The original API still works, but it's also possible to get the list of options without any need to modify the MountPoint object.

## Testing

With this change, the unit tests included in https://github.com/yast/yast-storage-ng/pull/995 just work.


storage/Filesystems/MountPoint.h
```diff
@@ -108,8 +108,15 @@ namespace storage
 	void set_mount_options(const std::vector<std::string>& mount_options);
 
 	/**
-	 * Set the mount options to the default mount options. So far the
-	 * default mount options only contain the subvol for btrfs subvolumes.
+	 * Default mount options. So far the default mount options only contain
+	 * the subvol for btrfs subvolumes.
+	 */
+	std::vector<std::string> default_mount_options() const;
```
aschnell: This could be called get_default_mount_options for consistency.

aschnell:
"break into pieces" sounds rather negative to me. Maybe "split ..." would be better.

storage/Filesystems/MountPoint.h
ancorgs: But I fear that having a method called `set_default_mount_options` and another called `get_default_mount_options` would give a false impression about how they work.

They would look as setter/getter for the same "attribute".

To be honest, I think that `MountableImpl#get_default_mount_options` should not include the `get_` prefix, it makes it look like a getter.

ancorgs: Just to clarify, I'm not opposing to change the name to `get_default_mount_options` (in the ruby wrapper it will look the same with or without the `get_`). I just explaining that I intentionally broke consistency with the internal names on purpose, because I think the result is less confusing.
As mentioned, in the yast-storage-ng side, it will make no difference.

aschnell: Then for consistency rename the internal function.

storage/Filesystems/MountPoint.cc
```diff
@@ -137,6 +137,13 @@ namespace storage
     }
 
 
+    vector<string>
+    MountPoint::default_mount_options() const
+    {
+	return get_impl().default_mount_options().get_opts();
```
joseivanlopez: NP: indentation looks wrong, doesn't it?

ancorgs: I think it's right, consistent to the rest of the file.

aschnell:
LGTM

storage/Filesystems/MountPoint.cc
joseivanlopez: Ok, github is not rendering it as I expected.

PR #678, opened 2019-11-20
Bind mount workaround

joseivanlopez:
### Problem

When probing a mount point for a mountable device, all possible candidate mount points are taken from both files: /etc/fstab and /proc/mounts. Then there is a "join" process, where fstab entries are associated to proc entries. This association is done by device name. So, all proc entries of a device are associated to a fstab entry of that device.

As result, when a directory is bind-mounted, last entry in  */proc/mounts* for a device is probed as its filesystem mount point.

For example:

~~~
$ cat /etc/fstab

/dev/sda1  /  ext4  defaults  0  0

$ mound /tmp/foo /mnt -o bind

$ cat /proc/mounts

/dev/sda1 / ext4 rw,relatime 0 0
/dev/sda1 /mnt ext4 rw,relatime 0 0
~~~

Here, */mnt* would be probed as mount point for the the filesystem over */dev/sda1*.

### Solution

Two fixes are provided to improve the mount point selection:

a) The association process has been modified to only associate a fstab entry with a proc one when both point to the same path (instead to the same device).

b) After the association process, the most reasonable mount point is selected. That means:

* shortest path from active fstab mount points, or
* shortest path from active mount points, or
* shortest path

Shortest path is selected because right now several mount points are not supported.

With these changes, */* (shortest active fstab mount point) would be probed as mount point of */dev/sda1* in the example above.

### Tests

* Unit tests have been added.
* Manually tested for EXT4 and for BTRFS including subvolumes.


aschnell:
LGTM

But seeing that the heuristic to select a mount point from the system was already fixed once before I wonder if we should not finally support multiple mount points per mountable. Something for the PO to think about.

joseivanlopez:
> LGTM
> 
> But seeing that the heuristic to select a mount point from the system was already fixed once before I wonder if we should not finally support multiple mount points per mountable. Something for the PO to think about.

Thanks! Yes, I agree. Maybe PO has it under his radar.

ancorgs:
I will create an epic card for supporting multiple mountpoints per mountable. But to be honest, it's not something I see happening in the short term. It likely implies rethinking (some parts of) the UI of the Partitioner and that kind of stuff.

On the other hand, I guess the future possibility of supporting it is something to take into account while redesigning the UI to implement proper support for btrfs subvolumes and all its features (something that is really in the agenda for the mid-term).

PR #679, opened 2019-11-26
improved calculation of resize information

aschnell:
For https://trello.com/c/uVzDozH4/1450-3-1149148-libstorage-ng-is-too-pessimistic-when-deciding-whether-is-possible-to-shrink-a-filesystem.

So far the library had a simple a pessimistic way to calculate the min size of filesystems which
was to high for some cases. Also the 64bit feature of ext4 was not used when calculating the
max size.

Now the calculation is more advanced. It does:

- use estimation from resize2fs for min size of ext2/3/4
- use 64bit feature for max size of ext4

Also the general calculation for the min size was improved. It is based on the size of the block device and values from statvfs now.


storage/Utils/StorageDefines.h
```diff
@@ -98,17 +98,17 @@
...
 #define RPCBINDBIN     "/sbin/rpcbind"
 
 #define EFIBOOTMGRBIN "/usr/sbin/efibootmgr"
 
-#define NTFSRESIZEBIN "/usr/sbin/ntfsresize"
+#define NTFSRESIZE_BIN "/usr/sbin/ntfsresize"
 #define XFSGROWFSBIN  "/usr/sbin/xfs_growfs"
 #define REISERFSRESIZEBIN "/sbin/resize_reiserfs"
-#define EXT2RESIZEBIN "/sbin/resize2fs"
+#define RESIZE2FS_BIN "/sbin/resize2fs"
```
shundhammer: NP: Use `/usr/sbin` instead of just `/sbin` for consistency?

On my (TW) system I have all those commands in `/usr/sbin` and just symlinks to them in `/sbin`.
I don't think that those symlinks will go away anytime soon, but still with all the other commands here using the `/usr/sbin` path it would only be consistent to use it here as well.

shundhammer:
Just the one NP issue with the `/usr/sbin` vs. `/sbin` path, but this really comes down to style and personal preferences.

Otherwise LGTM.

PR #680, opened 2019-11-28
- fixed escaping of graphviz escString (bsc#1157916)

aschnell:
In the graphviz file string that can contain a backslash (e.g. a subvolume path) must be escaped as a graphviz escString. Otherwise the backslash is not shown.

Part of https://trello.com/c/uv6c4Y1p/1333-2-p3-1135341-yast2-disk-coredump-problem-parsing-subvolume-name.

Also see https://bugzilla.suse.com/show_bug.cgi?id=1157916.


shundhammer:
LGTM

PR #681, opened 2019-11-29
- fixed handling of btrfs subvolumes with special (regex control) characters in the path (bsc#1135341)

aschnell:
For https://trello.com/c/uv6c4Y1p/1333-2-p3-1135341-yast2-disk-coredump-problem-parsing-subvolume-name and https://bugzilla.suse.com/show_bug.cgi?id=1135341.

The problem was that compilation of a regex failed when the btrfs subvolume path included regex control characters. To avoid that problem the characters are escaped now.

Unit tests are added. Manual tests have also been done.


wfeldt:
lgtm

PR #683, opened 2019-11-29
- extended unit test (bsc#1135341)

aschnell:
Extended one of the latest unit tests. Test is explained in the source code.

Also update documentation to reflect a recent change.

### DIFF
--- a/storage/Utils/AppUtil.cc
+++ b/storage/Utils/AppUtil.cc
@@ -477,6 +477,11 @@ namespace storage
     string
     regex_escape(const string& s, regex::flag_type f)
     {
+	// If the function is needed for any other flags it will have to be checked or
+	// even extended.
+	if (f != regex::extended)
+	    ST_THROW(Exception("regex_escape does not support used flags"));
+
 	// All special characters outside of any context, see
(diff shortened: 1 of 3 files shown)


testsuite/Utils/regex.cc
```diff
@@ -18,7 +18,9 @@ BOOST_AUTO_TEST_CASE(test_regex_escape1)
 
 /*
  * Check that for single char strings 1. the regex compilation with
- * the escaped string works and that 2. the string only matches itself.
+ * the escaped string works and that 2. the string only matches
+ * itself. The checks are also done when the escaped string is places
```
joseivanlopez: NP: typo s/places/placed

storage/Utils/AppUtil.cc
```diff
@@ -477,6 +477,11 @@ namespace storage
     string
     regex_escape(const string& s, regex::flag_type f)
     {
+	// If the function is needed for any other flags it will have to be checked or
+	// even extended.
+	if (f != regex::extended)
```
joseivanlopez: Would this require to bump the patch version?

aschnell: I do not think so since the function is only internal.

PR #684, opened 2019-12-03
show all block devices for action texts of btrfs subvolumes

aschnell:
For https://trello.com/c/SNWmNOzY/1489-1-sles15-sp2-p2-1152523-build-451-creation-of-subvolumes-is-displayed-as-assigned-always-to-1st-partition or https://bugzilla.suse.com/show_bug.cgi?id=1152523.

The main problem was that compound action texts for btrfs subvolumes did only contain one block device of the btrfs which, although irrelevant, confused some users. The texts include now all block devices including the sizes.

For some other compound action texts from Btrfs, LVM, MD, Bcache the code was improved to reuse the join function from BlkDeviceImpl.cc.

And finally the explanations for translators now always include the example sizes in the correct format.

### DIFF
--- a/storage/CompoundAction/Formatter.cc
+++ b/storage/CompoundAction/Formatter.cc
@@ -138,30 +138,4 @@ namespace storage
 	return first_action->text(commit_data);
     }
 
-
-    Text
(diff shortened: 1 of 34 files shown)


shundhammer:
LGTM

PR #687, opened 2019-12-10
- use correct size for btrfs resize command during shrink (bsc#1157762)

aschnell:
For https://trello.com/c/hCCPsANv/1492-5-p2-1157762-shrinking-btrfs-only-shrinks-partition-not-btrfs and https://bugzilla.suse.com/show_bug.cgi?id=1157762.

The problem was that the blk device used when getting the new size for the btrfs was the blk device of the probed devicegraph, while it has to be the one of the staging devicegraph. To be able to select the correct blk device in Btrfs::Impl::do_resize() the signature of the do_resize() functions was changed.

For consistency the signature of the do_resize_text(), do_reallot_text() and do_reallot() functions was also changed.

Several tests have been extended or added.


shundhammer:
LGTM

PR #688, opened 2019-12-10
Add specialized Luks callbacks

joseivanlopez:
### Problem

When activating LUKS devices, the callback method only provides the LUKS uuid. This makes very difficult to identify what device it refers when prompted to give the password.

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1125774

* Part of https://trello.com/c/MR24HhqL/1486-2-research-p4-1125774-sles-15-sp1-beta3-improve-naming-of-crypt-devices-in-the-passphrase-prompt

### Solution

Now on, a new specialized callback class is provided (`ActivateCallbacksLuks`). Clients can use this new callbacks when more information about the LUKS device is required.

Note that a new derived callbacks class is added to keep API/ABI backward compatibility. 

### Testing

* Manually tested.


aschnell:
That looks like an API change, not just and addition, so the major SO number would have to be increased with all the consequences.

storage/Storage.h
```diff
@@ -115,6 +115,64 @@ namespace storage
...
+
+	~LuksInfo();
+
+	std::string get_device_name() const;
+
+	std::string get_uuid() const;
+
+	std::string get_label() const;
+
+	void set_device_name(const std::string& device_name);
```
aschnell: These seem not required in the interface.

storage/Storage.h
```diff
@@ -115,6 +115,64 @@ namespace storage
...
+     */
+    class LuksInfo : private boost::noncopyable
+    {
+    public:
+
+	LuksInfo();
+
+	~LuksInfo();
+
+	std::string get_device_name() const;
```
aschnell: Could return const reference.

storage/Storage.h
joseivanlopez: Ok. I have added parameters to the constructor to avoid setters in the public interface.

joseivanlopez: Finally I have adapted it according to @aschnell suggestion (set the values by using the impl object directly). Thanks!

aschnell:
LGTM

PR #691, opened 2019-12-20
Partition Id names

joseivanlopez:
### Problem

Partition Id names are defined by the library, but the method is not exposed in the public API. This forces clients (i.e., YaST) to define their own partition id names.

Having two sources for partition id names is bad for some reason:

* Different names could be shown to the user in the UI.

* Translators have to translate both, libstorage-ng and YaST names.

Part of https://trello.com/c/QsLIaeYN/1519-tw-leap-151-storage-ng-inconsistent-partition-type-display

### Solution

Public API is extended with a new method `storage::get_partition_id_name`.

Tested manually.


storage/Devices/Partition.h
```diff
@@ -137,6 +137,12 @@ namespace storage
     };
 
 
+    /**
+     * Convert IdNum to string.
```
aschnell: The documentation should mention that the returned string can be empty.

aschnell:
LGTM

PR #693, opened 2020-01-08
Allow optional fstab columns

joseivanlopez:
### Problem

The last two columns of the *fstab* entries (fs_freq and fs_passno) are optional, but entries are being ignored if they lack of any of those options.

For example, if the user adds a line like the following to the *fstab*:

```
/home/user  /home/other/user  none  bind
```

The resulting *fstab* after the commit would not contain that line.

* https://bugzilla.suse.com/show_bug.cgi?id=1159557
* PBI: https://trello.com/c/BhPYesPV/1564-3-p2-1159557-yast-partitioner-removes-bind-entries-from-etc-fstab

### Solution

*fstab* parser has been improved to allow that optional columns (with default value to 0).


testsuite/fstab.cc
```diff
@@ -124,6 +124,54 @@ BOOST_AUTO_TEST_CASE( parse_and_format )
...
+    fstab.parse(input);
+
+    BOOST_CHECK_EQUAL(fstab.get_entry_count(), 2);
+
+    string_vec output = fstab.format_lines();
+
+    for (int i = 0; i < fstab.get_entry_count(); ++i)
+        BOOST_CHECK_EQUAL(output[i], expected_output[i]);
+
+    int i = 0;
```
aschnell: The use of ```i``` is a bit strange to me. Why not simply pass ```0``` and ```1```?

storage/EtcFstab.cc
```diff
@@ -286,8 +290,12 @@ namespace storage
             fs_type = FsType::UNKNOWN;
 
 	mount_opts.parse( EtcFstab::fstab_decode( get_column( col++ ) ), line_no );
-	dump_pass = atoi( get_column( col++ ).c_str() );
-	fsck_pass = atoi( get_column( col++ ).c_str() );
+
+	if ( column_count >= FSTAB_COLUMN_COUNT - 1 )
+	    dump_pass = atoi( get_column( col++ ).c_str() );
```
aschnell: ```dump_pass``` should be set to ```0``` if the value is not present. The ```parse``` function could be used on an entry that was not just created with clean values.

aschnell:
LGTM

PR #694, opened 2020-01-08
Fix partition id when creating partition

joseivanlopez:
### Problem

When a partition is created with id *DOS16* (0x06), the resulting partition has 0x0e id.

PBI: https://trello.com/c/QsLIaeYN/1519-tw-leap-151-storage-ng-inconsistent-partition-type-display

### Solution

In general, the partition id is set after creating the partition:

```
parted --script --wipesignatures /dev/sdb mkpart primary ext2 10MiB 1000MiB

parted --script /dev/sdb set 1 type 11
```

But parted is able to assign the correct partition id while creating the partition. For that, the proper filesystem has to be indicated:

```
parted --script --wipesignatures /dev/sdb mkpart primary linux-swap 10MiB 1000MiB
```

The command above creates the partition with id 0x82 (swap)

*Parted* used to use *fat16* filesystem to create a new partition with *DOS16* (0x06) id, but that does not work anymore. Now, the partition is created with id 0x0c when *fat16* is used.

This PR fixes the problem by performing the two steps when creating a partition with *DOS16* id: a) create the partition, and b) set the partition id.


aschnell:
LGTM

PR #698, opened 2020-01-14
- use udevadm from /usr/bin instead of /sbin

aschnell:
The bigger part of this PR is to use udevadm from /usr/bin instead of /sbin since the latter disappears.
The smaller part of this PR is to remove the option to enable crc from XFS from mkfs.xfs since it is the default for years anyway.

### DIFF
--- a/storage/Filesystems/XfsImpl.cc
+++ b/storage/Filesystems/XfsImpl.cc
@@ -88,7 +88,7 @@ namespace storage
     {
 	const BlkDevice* blk_device = get_blk_device();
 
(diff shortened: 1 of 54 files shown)


shundhammer:
LGTM

wfeldt:
Please note that there are yast-storage-ng test cases checking for `/sbin/udevadm`...

aschnell:
Those test cases deal with the code in yast2-storage-ng - which of course also needs to be adapted.

ancorgs:
See https://trello.com/c/5MQmiuTs/3829-ostumbleweed-p5-1160890-stop-using-sbin-udevadm-symlink

PR #699, opened 2020-01-14
Fix mount and unmount

joseivanlopez:
## Problem

When a mount point is edited and its mount path is modified, the unmount action is always performed, even when the probed mount point is inactive. This leads to an unmounting error message during the commit phase.

A similar problem happens when the mount path changes and the mount point is marked as inactive. Independently it was set as inactive, the device is mounted anyway.

These issues were detected when testing changes for https://trello.com/c/Z8gqtcC1/1484-3-nfs-client-and-unmounting-bug1006815-bug1151426.

See https://bugzilla.suse.com/show_bug.cgi?id=1151426.

### Solution

Unmount action is added only if the probed mount point is active, and the mount action is added only if the target mount point is active.


storage/Filesystems/MountPointImpl.cc
```diff
@@ -386,7 +386,7 @@ namespace storage
 	    }
 	}
 
-	if ((!lhs.active && active) || (lhs.path != path))
+	if (active && (!lhs.active || lhs.path != path))
```
jreidinger: well, as it is obviously due to previous bug not trivial condition, I would like to see comment here explaining when exactly it mount/unmount.

joseivanlopez: In fact, I think I have found a bug while documenting it :(

PR #703, opened 2020-02-04
Remove OutputProcessor class

joseivanlopez:
`OutputProcessor` was used for visual feedback to the user (progress bar, busy animation). But this is now done by clients (e.g., YaST).


LIBVERSION
```diff
@@ -1 +1 @@
-1.26.0
+1.26.1
```
aschnell: For this purely internal cleanup a version increase is not needed AFAIS.

aschnell:
LGTM

PR #707, opened 2020-03-03
Ignore clustered LVM

joseivanlopez:
### Problem

LVM tools provided by SLE-15-SP1 fails with error code 5 when it detects clustered LVM but clustered locking is not available. This makes to *libstorage-ng* to not probe other LVs in the system.

* https://bugzilla.suse.com/show_bug.cgi?id=1161775
* https://trello.com/c/PB0cvnjU/1670-2-bug1161775-clustered-lvms-fix-for-sle-15-sp1

### Solution

Option *--ignoreskippedcluster* is passed to the lvm commands (i.e., *lvs*, *pvs* and *vgs* commands):

~~~
 --ignoreskippedcluster
    Use to avoid exiting with an non-zero status code if the command is run without clustered locking and clustered VGs are skipped.
~~~

### Testing

* Adapted unit tests
* Manual testing was not done because I cannot emulate a clustered LVM.


wfeldt:
lgtm

joseivanlopez:
SR: https://build.suse.de/request/show/213273

dgdavid:
Just for the record

> Manual testing was not done because I cannot emulate a clustered LVM.

It seems that it's not needed to setup an LVM cluster for manual testing.

Here is what @wfeldt did to test it in https://github.com/openSUSE/libstorage-ng/pull/717

> ## Testing
> 
> You don't need to setup an LVM cluster to test this. It's sufficient to create a clustered volume group:
> 
> - edit `/etc/lvm/lvm.conf` to change `locking_type` to 3
> - create a clustered VG: `vgcreate -cy foo_vg /dev/foo`
> - revert `lvm.conf` change
> 
> That's it - since we don't do anything with that VG.

PR #708, opened 2020-03-05
Sync branches (SLE-15-SP1)

joseivanlopez:
Syncing SLE-15-SP1 and master branches.

Note that fixes in https://github.com/openSUSE/libstorage-ng/pull/707 only applies to SLE-15-SP1. Other solution will be proposed for master.


wfeldt:
fine with me

PR #709, opened 2020-03-12
track device block size when creating an Md RAID (bsc#1164295)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1164295
- https://trello.com/c/NIIfPuhi

A newly created RAID has the default block size of 512 bytes. But this might be wrong if the
devices it constitutes of have different sizes.

## Solution

When adding devices to a RAID, adjust RAID block size to the maximum block size of its disk devices.

## Notes

- there seems to be no documentation about this anywhere
- it was tested with different RAIDs and various setups - and the behavior is consistent
- mdadm will not complain about misaligned setups like range starts and sizes that can't be represented with the final block size - you get i/o errors from the kernel when accessing that device later
- it seems everyone just expects the admin to do the sensible thing - so I would not overengineer this on our side
- if there's need to investigate mdadm behavior more deeply (like do we really need to waste 128 MiB) I would go for a separate research PBI

### DIFF
--- a/bindings/storage-catches.i
+++ b/bindings/storage-catches.i
@@ -311,6 +311,7 @@
 %catches(storage::Exception) storage::Partitionable::get_default_partition_table_type() const;
 %catches(storage::WrongNumberOfChildren, storage::DeviceHasWrongType) storage::Partitionable::get_partition_table();
 %catches(storage::WrongNumberOfChildren, storage::DeviceHasWrongType) storage::Partitionable::get_partition_table() const;
+%catches(storage::InvalidBlockSize) storage::Region::adjust_block_size(unsigned int block_size);
 %catches(storage::Exception) storage::Region::adjust_length(long long delta);
 %catches(storage::Exception) storage::Region::adjust_start(long long delta);
 %catches(storage::Exception) storage::Region::get_end() const;
--- a/storage/Devices/MdImpl.cc
+++ b/storage/Devices/MdImpl.cc
@@ -830,9 +830,13 @@ namespace storage
 	unsigned long long sum = 0;
 	unsigned long long smallest = std::numeric_limits<unsigned long long>::max();
 
+	unsigned int block_size = 0;
+
 	for (const BlkDevice* blk_device : devices)
 	{
 	    unsigned long long size = blk_device->get_size();
+
+	    block_size = std::max( block_size, blk_device->get_region().get_block_size() );
 
 	    const MdUser* md_user = blk_device->get_impl().get_single_out_holder_of_type<const MdUser>();
 	    bool spare = md_user->is_spare();
@@ -913,6 +917,14 @@ namespace storage
 	    case MdLevel::CONTAINER:
 	    case MdLevel::UNKNOWN:
 		break;
+	}
+
+	// adjust block size
+	if (block_size && block_size != get_region().get_block_size())
+	{
+	    Region region(get_region());
+	    region.adjust_block_size(block_size);
+	    set_region(region);
 	}
 
 	set_size(size);
--- a/storage/Utils/RegionImpl.cc
+++ b/storage/Utils/RegionImpl.cc
@@ -107,6 +107,27 @@ namespace storage
     }
 
 
+    void
+    Region::Impl::adjust_block_size(unsigned int block_size)
+    {
+	assert_valid_block_size(block_size);
+
+	if (block_size == Impl::block_size)
+	    return;
+
+	if (start * Impl::block_size % block_size ||
+	    length * Impl::block_size % block_size)
+	{
+	    ST_THROW(InvalidBlockSize(block_size));
+	}
+
+	start = start * Impl::block_size / block_size;
+	length = length * Impl::block_size / block_size;
+
+	Impl::block_size = block_size;
+    }
+
+
     unsigned long long
     Region::Impl::to_bytes(unsigned long long blocks) const
     {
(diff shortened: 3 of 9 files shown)


storage/Devices/MdImpl.cc
```diff
@@ -830,10 +830,14 @@ namespace storage
 	unsigned long long sum = 0;
 	unsigned long long smallest = std::numeric_limits<unsigned long long>::max();
 
+        unsigned int block_size = 0;
```
ancorgs: Indentation looks wrong, in github at least.

storage/Devices/MdImpl.cc
```diff
@@ -830,10 +830,14 @@ namespace storage
...
+
 	for (const BlkDevice* blk_device : devices)
 	{
 	    unsigned long long size = blk_device->get_size();
 
+            block_size = std::max( block_size, blk_device->get_region().get_block_size() );
```
ancorgs: Same indentation problem.

storage/Devices/MdImpl.cc
wfeldt: Yeah, that's that tab/space mixture in the sources. :-(

joseivanlopez:
Good catch! Thanks.

storage/Devices/MdImpl.cc
```diff
@@ -915,6 +919,14 @@ namespace storage
 		break;
 	}
 
+	// adjust block size
+	if (block_size && block_size != get_region().get_block_size())
+	{
+	    Region region = Region(get_region());
```
jreidinger: why not simple `Region region(get_region());` ?

wfeldt: That would also do; I was probably just in a typing mood.

storage/Devices/MdImpl.cc
```diff
@@ -915,6 +919,14 @@ namespace storage
...
+	    region.adjust_block_size(block_size);
+	    set_region(region);
+	}
```
ancorgs: We would need to document the reasons for this.

I mean, I understand what the code implies - when creating a RAID, such RAID will have the bigger block size of all its underlying devices. But I miss a link to some documentation about it, maybe from `mdadm` or whatever.

Some place where I can get reply to questions such as:

 - Can I actually combine devices with different block sizes for all levels of RAID (mirroring, stripping, combinations)?
 - The biggest block size will always be used for the RAID in all cases (for all levels)?
 - What happens if some of the devices has a block size that is not divisible by the biggest one? Will mdadm fail? Will mdadm accept that but the result would be unreliable?

storage/Utils/RegionImpl.cc
```diff
@@ -107,6 +107,27 @@ namespace storage
...
+	assert_valid_block_size(block_size);
+
+	if (block_size == Impl::block_size)
+	    return;
+
+	if (start * Impl::block_size % block_size ||
+	    length * Impl::block_size % block_size)
+	{
+	    ST_THROW(InvalidBlockSize(block_size));
+	}
```
ancorgs: For the Region class, raising this exception is ok.

But in my opinion, we should have some previous check (in libstorage-ng or in yast2-storage-ng) to prevent the combination of incompatible devices into a RAID, either caching this exception or doing the check beforehand. Any plan/idea about it?

storage/Devices/MdImpl.cc
wfeldt: (1) there seems to be no problem; I've experimented a bit
(2) it makes sense but I doubt it's documented somewhere; we'd have to check the kernel sources
(3) that's impossible as block sizes are always powers of 2

storage/Utils/RegionImpl.cc
wfeldt: That's really a fringe case and can happen if you combine an odd-sized 0.5k disk with a 4k disk into a raid.

I'm fine with throwing a different exception if you have any suggestion. So that you don't start to replicate the logic in y2storage.

Not sure that this is still in scope of the pr, which is about fixing a bug.

storage/Devices/MdImpl.cc
ancorgs: > (3) that's impossible as block sizes are always powers of 2

Right. I didn't expressed it correctly, but I was actually refferring to the fringe case you described in the other comment. What happens in that case from the mdadm POV? Will it fail, will it create an unreliable device? My expectation was actually a device that works but is one block smaller.

wfeldt: See 'Notes' above

bindings/storage-catches.i
```diff
@@ -311,6 +311,7 @@
 %catches(storage::Exception) storage::Partitionable::get_default_partition_table_type() const;
 %catches(storage::WrongNumberOfChildren, storage::DeviceHasWrongType) storage::Partitionable::get_partition_table();
 %catches(storage::WrongNumberOfChildren, storage::DeviceHasWrongType) storage::Partitionable::get_partition_table() const;
+%catches(storage::InvalidBlockSize) storage::Region::adjust_block_size(unsigned int block_size);
```
ancorgs: Honest question (I don't know how this work): is it enough for the exception to be propagated?

I mean, in the Ruby code we will likely never call `Region::adjust_block_size` directly. We will actually see the exception as a consequence of `Md::add_device`. I hope SWIG takes care of those details, but better to ask :-)

wfeldt: Honestly, I donät know all the SWIG details. But it worked when I tested it from the srorage-ng side.

PR #712, opened 2020-03-20
update doc on RAID block size

wfeldt:
## Task

- https://trello.com/c/eixP9hzP

Update doc on RAID block size.

### DIFF
--- a/doc/md-raid.md
+++ b/doc/md-raid.md
@@ -88,3 +88,18 @@ MdMembers can be used as generic Mds.
 
 MdContainers and MdMembers can be added to and removed from /etc/mdadm.conf.
 
+Combining disks with different block sizes
+------------------------------------------
+
+If you combine disks with different block sizes into a RAID, the RAID device
+will have the maximum block size of its disks.
+
+This seems to work (at least most of the time) even if some parts of the
+RAID are not aligned to this block size.
+
+Md::add_device() takes care to update Md::Region to use the correct block
+size. It is, however not possible to combine regions with different block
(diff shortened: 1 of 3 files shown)


storage/Devices/MdImpl.cc
```diff
@@ -805,6 +805,8 @@ namespace storage
 	// big can lead to severe problems later on, e.g. a partition not
 	// fitting anymore, we make a conservative calculation.
 
+        // For combining disks with different block sizes, see doc/md-raid.md.
```
ancorgs: Mixed indentation strikes back

doc/md-raid.md
```diff
@@ -88,3 +88,18 @@ MdMembers can be used as generic Mds.
...
+
+This seems to work (at least most of the time) even if some parts of the
+RAID are not aligned to this block size.
+
+Md::add_device() takes care to update Md::Region to use the correct block
+size. It is, however not possible to combine regions with different block
+sizes (in libstorage-ng) if the start or length of a region are misaligned.
+
+This is, however, not really a problem here as the region start is always 0
+and the RAID size is an approximation anyway - so we might simply round down.
```
ancorgs: Rounding down sounds reasonable to me

ancorgs: Could we stop raising the `InvalidBlockSize` If we implement that rounding down?

wfeldt: Yes. But I'd like not to mix everything up. This is only the doc.Then change the code. Then the backport.

PR #713, opened 2020-03-25
Make BlkDevice::possible_mount_bys public (for bsc#1166096)

ancorgs:
To fix [bsc#1166096](https://bugzilla.suse.com/show_bug.cgi?id=1166096), we need to extend a bit the API of yast2-storage-ng, allowing to get the preferred udev name of block devices that are not formatted. See https://github.com/yast/yast-storage-ng/pull/1069

It would be nice to not duplicate in yast2-storage-ng the logic that is already implemented in libstorage-ng as `BlkDevice::possible_mount_bys`. This PR makes that function public to avoid that duplication.


ancorgs:
See this comment https://github.com/yast/yast-storage-ng/pull/1069#discussion_r397662613

wfeldt:
lgtm

PR #718, opened 2020-04-01
respect number of partition slots reported by parted for GPT

aschnell:
We have a report (https://bugzilla.suse.com/show_bug.cgi?id=1161783) where a special GPT has only 15 partition slots. So far libstorage-ng assumed 128 partition slots.

parted was extended to report the number of partition slots. With this patch libstorage-ng queries that number and handles it (only for GPTs).


shundhammer:
LGTM

PR #719, opened 2020-04-16
document device block size handling

wfeldt:


doc/blocksizes.md
```diff
@@ -0,0 +1,296 @@
...
+[ 1018.710597] sd 0:0:5:0: [sdf] tag#37 request not aligned to the logical block size
+[ 1018.710650] blk_update_request: I/O error, dev sdf, sector 2107391 op 0x0:(READ) flags 0x1000 phys_seg 1 prio class 0
+```
+
+### LVM Issues
+
+- Can you reliably know in advance which PVs will be used? Probably not.
+
+- Maybe read block size from system after creating the LVs; but at least
+  before using parted.
```
aschnell, line 128: parted is not used on logical volumes.

wfeldt: Good point.

aschnell:
LGTM

PR #720, opened 2020-04-17
fix unit test

wfeldt:
## Task

A unit test fails when `/etc/fstab` is missing.

## Solution

`/etc/fstab` may be missing in docker image. Use a file that's guaranteed to exist and not have the exec bit set.


aschnell:
LGTM

PR #721, opened 2020-04-20
add support to probe lvm cache

aschnell:
For https://trello.com/c/D8BBQ5S3/280-8-ostumbleweed-p3-1099744-follow-up-probe-lvmcache.

Add support to probe lvm cache devices. Documentation and unit tests are included.


storage/Devicegraph.h
```diff
@@ -199,32 +213,89 @@ namespace storage
 	 */
 	bool holder_exists(sid_t source_sid, sid_t target_sid) const;
 
+	/**
+	 * Clear the devicegraph.
```
shundhammer, line 217: Maybe mention that this invalidates all pointers to objects in the devicegraph?
It _should_ be clear, but we had bugs about this before...

storage/Devicegraph.h
```diff
@@ -199,32 +213,89 @@ namespace storage
...
+	 */
 	void clear();
 
-	// convenient functions, equivalent to e.g. Disk::get_all(devicegraph)
+	/**
+	 * Get all Disks.
+	 *
+	 * Convenient functions, equivalent to Disk::get_all(devicegraph).
```
shundhammer, line 224: _Convenience_ functions (search and replace globall)

storage/Devices/BlkDevice.h
```diff
@@ -86,7 +86,24 @@ namespace storage
 	const std::string& get_sysfs_name() const;
 	const std::string& get_sysfs_path() const;
 
+	/**
+	 * Return the names of the udev by-path links of the blk
+	 * device. The names do not include the directory.
```
shundhammer, line 91: > The names do not include the directory

An example or two might help here. So what do the names look like?

storage/Devices/Encryption.h
```diff
@@ -170,7 +170,14 @@ namespace storage
 	 */
 	void set_open_options(const std::string& open_options);
 
+	/**
+	 * Get all Encryptions.
```
shundhammer, line 174: Maybe mention that encryption can be stacked in multiple layers upon one another?

Also, does this return that stack, or only the _different_ encryption types (like `sort | uniq`)?

storage/Devices/LvmLv.h
```diff
@@ -74,18 +132,29 @@ namespace storage
 	static std::vector<const LvmLv*> get_all(const Devicegraph* devicegraph);
 
 	/**
-	 * Get logical volume name. This is different from get_name().
+	 * Get the logical volume name. This is different from get_name().
```
shundhammer, line 135: Please elaborate. Which is which? How are they different? Examples?

storage/Devices/LvmLvImpl.cc
```diff
@@ -120,7 +120,8 @@ namespace storage
     bool
     LvmLv::Impl::is_usable_as_blk_device() const
     {
-	return lv_type == LvType::NORMAL || lv_type == LvType::THIN || lv_type == LvType::RAID;
+	return lv_type == LvType::NORMAL || lv_type == LvType::THIN || lv_type == LvType::RAID ||
```
shundhammer, line 123: This would be a good candidate for a `switch` and enumerating all possible values; that way the compiler would complain if a new enum value is added, yet forgotten here.

```C++
switch (lv_type)
{
    case LvType:NORMAL:
    case LvType::THIN:
    ...
        return true;

    case LvType::WHATEVER:
    case LvType::WHATNOT:
        return false;
}
```

storage/Devices/LvmLvImpl.cc
```diff
@@ -245,33 +250,67 @@ namespace storage
...
+		case LvType::CACHE:
+		case LvType::WRITECACHE:
 		{
-		    LvmVg* lvm_vg = LvmVg::Impl::find_by_uuid(prober.get_system(), lv.vg_uuid);
-		    lvm_lv = lvm_vg->create_lvm_lv(lv.lv_name, lv.lv_type, lv.size);
+		    // A cache is private if uses as a thin-pool.
+
+		    if (lv.role == CmdLvs::Role::PUBLIC)
+		    {
+			LvmVg* lvm_vg = LvmVg::Impl::find_by_uuid(system, lv.vg_uuid);
```
shundhammer, line 261: There appears to be an awful lot of code duplication here; this sequence repeats 4 times.

This calls for factoring out into a separate function that accepts the two parameters that are different: The UUID to find and the LvType to create.

storage/Devices/Partitionable.h
```diff
@@ -43,7 +43,17 @@ namespace storage
 	const Topology& get_topology() const; // deprecated since also available in BlkDevice
 	void set_topology(const Topology& topology); // deprecated since also available in BlkDevice
 
+	/**
+	 * Query the range of device nodes available.
```
shundhammer, line 47: Please elaborate what this is. What is _range_ in this context?

storage/Filesystems/MountPoint.h
```diff
@@ -144,7 +151,7 @@ namespace storage
 	/**
 	 * Set the freq field.
```
shundhammer, line 152: Please mention that this refers to the fsck pass.

storage/Filesystems/MountPoint.h
```diff
@@ -144,7 +151,7 @@ namespace storage
...
 	 *
```
shundhammer, line 153: Please mention that this is the _dump frequency_ of the utterly obsolete _dump_ backup tool from very early Unix versions and that this has zero relevance today.

storage/Holders/Holder.h
```diff
@@ -57,11 +57,21 @@ namespace storage
 	Device* get_source();
```
shundhammer, line 57: This header file is sorely missing the explanation what a _holder_ is.
It is simple when you had any graph theory during studying; it's the edges connecting nodes. But this is not clear unless one knows the lingo of the underlying Boost graph library.

storage/Devices/LvmLvImpl.cc
aschnell: Unfortunately it is a bit more different. For thins the parent of the logical volume is not the volume group but the thin pool, so an object of a different type. So at least two helper functions would be needed.

shundhammer: OK then.

shundhammer:
LGTM

storage/Devicegraph.h
```diff
@@ -118,7 +118,14 @@ namespace storage
 
 
     /**
-     * The master container of the libstorage.
+     * The master container of the libstorage-ng.
+     *
+     * The devicegraph contains Devices (node) and Holders (edge). A
+     * device represent a storage object, e.g. a disk, a LVM volume
+     * group, a file systems or a mount point. A holder represent the
```
shundhammer, line 125: represent**s**

storage/Holders/Holder.h
```diff
@@ -46,8 +46,12 @@ namespace storage
     };
 
 
-    // abstract class
-
+    /**
+     * An abstract base class for storage holders. Storage holders are
+     * edges in the devicegraph.
```
shundhammer, line 51: :+1:

storage/Devicegraph.h
```diff
@@ -118,7 +118,14 @@ namespace storage
 
 
     /**
-     * The master container of the libstorage.
+     * The master container of the libstorage-ng.
+     *
+     * The devicegraph contains Devices (node) and Holders (edge). A
```
shundhammer, line 123: :+1:

aschnell:
Somehow Travis CI is stuck: One job is running for more than an hour now and the later once are not even started. I can only ignore these temporary Travis CI problems.

PR #723, opened 2020-04-22
set default to false for in_etc_crypttab() (bsc#1118977)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1118977
- https://trello.com/c/jtojel4R

New encrypted volumes always show up in `/etc/crypttab` - even when they should not.

## Solution

The default should be `false` to not have enrypted volumes show up unexpectedly in `/etc/crypttab`.

yast-storage-ng will change the attribute as needed.

## See also

- https://github.com/yast/yast-storage-ng/pull/1087

### DIFF
--- a/storage/Devices/EncryptionImpl.cc
+++ b/storage/Devices/EncryptionImpl.cc
@@ -51,7 +51,7 @@ namespace storage
 
     Encryption::Impl::Impl(const string& dm_table_name)
 	: BlkDevice::Impl(DEV_MAPPER_DIR "/" + dm_table_name), type(EncryptionType::LUKS1), password(),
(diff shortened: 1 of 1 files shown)


aschnell:
Changing such a default value looks dangerous to me. It might break existing code. It also seems inconsistent with in_etc_fstab flag - which is also true.

AFAIS integration-tests/luks/create-with-ext4.py will now create the entry in fstab but not crypttab an booting will fail.

ancorgs:
I'm also not sure. libstorage-ng should work properly on its own. It looks weird to rely on yast2-storage-ng to adjust something that may have been initially set up correctly or incorrectly by the library.

wfeldt:
ok, let's dump this - but see https://github.com/yast/yast-storage-ng/pull/1087 update

PR #725, opened 2020-04-28
- added read-only flag to BlkDevice

aschnell:
Add a read-only flag to block devices. E.g. LVM logical volumes can be read-only,
and an LVM external origin must be.


shundhammer:
LGTM

PR #727, opened 2020-04-30
- work on probing lvm snapshots

aschnell:
This PR adds support to probe LVM snapshots (both think and thin). It does not yet probe the relationship between origin and snapshot. That will come later.


wfeldt:
lgtm

PR #736, opened 2020-05-08
- added sanity checks

aschnell:
Added some sanity checks when looking for btrfs subvolumes during probing. Without the new checks unspecific error would happen.

### DIFF
--- a/storage/SystemInfo/CmdBtrfs.cc
+++ b/storage/SystemInfo/CmdBtrfs.cc
@@ -182,9 +182,13 @@ namespace storage
 	    Entry entry;
 
 	    string::size_type pos1 = line.find("ID ");
+	    if (pos1 == string::npos)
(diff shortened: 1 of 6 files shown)


storage/SystemInfo/CmdBtrfs.cc
```diff
@@ -182,9 +182,13 @@ namespace storage
 	    Entry entry;
 
 	    string::size_type pos1 = line.find("ID ");
+	    if (pos1 == string::npos)
+		ST_THROW(Exception("cound not find id"));
```
wfeldt, line 186: (in all three places): 'could'

BTW, could that message be a bit more descriptive, like 'could not find id in btrfs output' or so?

aschnell: Sure.

wfeldt:
lgtm

PR #738, opened 2020-05-11
- allow to probe btrfs snapshot relations

aschnell:
With this PR libstorage-ng is able to probe btrfs snapshpt relations, that is what subvolume is a snapshot of what subvolume. As a first commit it changes the devicegraph definition to allow parallel edges that can happen now.

There is still a problem: It is possible to create a setup with a cycle in the devicegraph (see cycle1 test). That can also cause the actiongraph to have a cycle.

This devicegraph shows the new snapshot relations in green for the two new cases (parallel edge and cycle):

btrfs

For the time being the new probing is only active with YAST_BTRFS_SNAPSHOT_RELATIONS=yes.

Not for any specific bug, feature or PBI.


storage/Devicegraph.h
```diff
@@ -325,15 +334,28 @@ namespace storage
 	void remove_devices(std::vector<Device*> devices);
 
 	/**
-	 * @throw HolderNotFoundBySids
+	 * Find a holders with source_sid and sid_t target_sid. Fails
```
shundhammer, line 337: "Find holders" or "Find a holder"

storage/EnvironmentImpl.cc
```diff
@@ -109,4 +109,12 @@ namespace storage
 	return !p || strcmp(p, "yes") == 0;
     }
 
+
+    bool
+    support_btrfs_snapshot_relations()
+    {
+	const char* p = getenv("YAST_BTRFS_SNAPSHOT_RELATIONS");
```
shundhammer, line 116: Any idea how often this function is called? A couple times per program run? A couple hundred times? More?

If it's called very often, would it make sense to cache this value and check the environment only the first time?

storage/SystemInfo/CmdBtrfs.cc
```diff
@@ -203,6 +203,18 @@ namespace storage
 	    if (boost::starts_with(entry.path, "<FS_TREE>/"))
 		entry.path.erase(0, strlen("<FS_TREE>/"));
 
+	    string::size_type pos4 = line.find(" uuid ");
+	    if (pos4 == string::npos)
+		ST_THROW(Exception("could not find 'uuid' in 'btrfs subvolume list' output"));
+	    line.substr(pos4 + 6) >> entry.uuid;
```
shundhammer, line 209: How about `sizeof("somestring") - 1` for all those numbers to get an idea what we are skipping over?

storage/EnvironmentImpl.cc
aschnell: Right now once per btrfs snapshot during probing. Given the fact that also e.g. lsattr is called for every subvolume the call of getenv seems insignificant. Also when these snapshot relations are stable the whole function can be removed.

storage/SystemInfo/CmdBtrfs.cc
aschnell: Sure.

storage/EnvironmentImpl.cc
shundhammer: OK.

shundhammer:
LGTM

PR #739, opened 2020-05-12
- disable more code if probing btrfs snapshot relations is disabled

aschnell:
There are always surprises. In this case the output of 'btrfs subvolume show' does not include a UUID.

https://openqa.opensuse.org/tests/1263632#step/welcome/6


shundhammer:
LGTM

PR #740, opened 2020-05-13
- ignore some snapshot relations in actiongraph

aschnell:
Added new View::REMOVE. From the inline documentation:

> A special view for removing devices. Identical to CLASSIC except that holders of type Snapshot are visible when the source is a normal LVM logical volume.
>
> This view is useful when removing descendants. When removing a normal logical volume with snapshots the snapshots must also be removed. When removing a thin logical volume the snapshots can stay. This view implements this behaviour.

The new view is used when calculating dependencies for removing devices (in the actiongraph).

Also views are now visible in the public interface and some functions accept a view parameter.


shundhammer:
LGTM

PR #741, opened 2020-05-13
- use views in actiongraph calculation

aschnell:
Some code was changed to use functions like ```get_parents()``` instead of ```inv_adjacent_vertices()``` with ```boost::tie``` which should be easier to understand. The new code also allows to use the new views.


shundhammer:
LGTM

PR #742, opened 2020-05-14
- moved load registry to separate source file

aschnell:
Split registry of load function to separate file to reduce size of original source file.

Not for any specific PBI, bug or feature.


shundhammer:
LGTM

PR #744, opened 2020-05-18
- disallow resizing thick volume groups with snapshots

aschnell:
Resizing non-thin logical volumes with snapshots is problematic since the logical volume and all snapshots must be inactive. For the time being resizing is disallowed.

yast2-storage-ng will need a new message for the new enum value.

### DIFF
--- a/storage/Holders/Snapshot.h
+++ b/storage/Holders/Snapshot.h
@@ -31,13 +31,8 @@ namespace storage
 {
 
     /**
-     * Relationship between snapshot and snapshot. The snapshot is the
-     * source and the snapshot the target. So far only used for LVM
-     * snapshot.
-     *
-     * Could also be used for btrfs snapshots.
-     *
-     * So far internal to the library.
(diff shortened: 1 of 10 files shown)


storage/Holders/Snapshot.h
```diff
@@ -31,13 +31,8 @@ namespace storage
...
 
     /**
-     * Relationship between snapshot and snapshot. The snapshot is the
-     * source and the snapshot the target. So far only used for LVM
-     * snapshot.
-     *
-     * Could also be used for btrfs snapshots.
-     *
-     * So far internal to the library.
+     * Relationship between snapshot and snapshot. The snapshot is the source and the
```
shundhammer, line 34: This is confusing. The snapshot is both the source and the target? Or shouldn't one of those be something else than "snapshot"?

shundhammer:
Besides the confusing comment, LGTM.

shundhammer:
LGTM now.

PR #745, opened 2020-05-19
- correct block device size for LVM thick snapshots

aschnell:
This PR makes the block device size of LVM thick snapshots correct. The lvs command reports (as 'lv_size') the size allocated for differences to the origin - so not the block device size. The block device size is also reported by lvs as the 'origin_size').

So now also 'origin_size' is queried from lvs and the new variable ```used_extents``` is added to LvmLv::Impl to keep track of the allocated size for snapshots.


shundhammer:
LGTM

PR #751, opened 2020-06-19
- work on pools

aschnell:
This PR adds the capability to query the max partition size for pools.

It also improves the program to generate the exceptions-file for swig to do some checks.

Not for any specific PBI, bug or feature.

### DIFF
--- a/storage/Pool.h
+++ b/storage/Pool.h
@@ -109,7 +109,7 @@ namespace storage
 	std::vector<const Device*> get_devices(const Devicegraph* devicegraph) const;
 
 	/**
-	 * Find the maximal partition size the pool can provide for the given number of
+	 * Find the maximum partition size the pool can provide for the given number of
 	 * partitions.
 	 *
 	 * @throw PoolOutOfSpace, Exception
--- a/storage/Pool.h
+++ b/storage/Pool.h
@@ -115,7 +115,7 @@ namespace storage
 	 *
 	 * The result is nondeterministic.
 	 *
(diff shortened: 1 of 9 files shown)


storage/Pool.h
```diff
@@ -106,16 +108,25 @@ namespace storage
 	 */
 	std::vector<const Device*> get_devices(const Devicegraph* devicegraph) const;
 
+	/**
+	 * Find the maximal partition size the pool can provide for the given number of
```
shundhammer, line 112: s/maximal/maximum

storage/PoolImpl.cc
```diff
@@ -118,11 +118,64 @@ namespace storage
...
+	if (number == 0)
+	    ST_THROW(Exception("illegal value of zero for number"));
+
+	// For each partitionable find the biggest unused region.
+
+	vector<unsigned long long> candidates;
+
+	for (Partitionable* partitionable : get_partitionable_candidates(devicegraph))
+	{
+	    PartitionTable* partition_table = partitionable->get_partition_table();
```
shundhammer, line 135: Can this return NULL for a disk without a partition table (but maybe a filesystem directly on it)?
In that case, a check is missing to prevent a segfault.

If `get_partitionable_candidates()` does not return such disks, nevermind (but then a comment might be useful mentioning that).

aschnell: Added a note.

storage/Pool.h
aschnell: Fixed.

PR #754, opened 2020-06-29
- added UsedFeature for NVMe disks (for bsc#1172866)

aschnell:
This PR adds a UsedFeature for NVMe disks.

For https://trello.com/c/QIyuWNgI/1911-5-sles15-sp2-p2-1172866-nvme-cli-not-installed-when-root-is-an-nvme-device and https://bugzilla.suse.com/show_bug.cgi?id=1172866.

PR #757, opened 2020-06-29
- also add a UsedFeature for pmem devices

aschnell:
Similar to https://github.com/openSUSE/libstorage-ng/pull/754 also add a UsedFeature for PMEM devices.

For for any PBI, bug or feature.


shundhammer:
LGTM

storage/UsedFeatures.h
```diff
@@ -61,6 +61,7 @@ namespace storage
 	UF_FCOE = 1 << 18,
 	UF_FC = 1 << 19,
 	UF_DASD = 1 << 20,
+	UF_PMEM = 1 << 29,
```
kobliha, line 64: Hmm, maybe nitpicking, but it seems that this should be sorted somehow better. Extending this more and more can cause errors.

aschnell: It is sorted into categories.

PR #759, opened 2020-07-06
experimental support for RAID journal device

aschnell:
This PR adds experimental support for RAID journal device. A journal device can be used for RAID levels 4, 5, and 6. In write-through mode the journal device is used to avoid the infamous write-hole. In write-back mode it can also improve speed.

Since a journal device does not work for metadata version 1.0 it is not also possible to explicitly set another metadata version. This is also experimental.


storage/Holders/MdUserImpl.h
```diff
@@ -60,18 +60,29 @@ namespace storage
...
 
+	bool is_journal() const { return journal; }
+	void set_journal(bool journal);
+
 	unsigned int get_sort_key() const { return sort_key; }
 	void set_sort_key(unsigned int sort_key) { Impl::sort_key = sort_key; }
 
     private:
 
+	// TODO is an enum instead of three booleans more appropriate?
```
shundhammer, line 71: Using an enum here would either mean that those flags are absolutely mutually exclusive, or it would become a OR'ed single field (which would mean the checks would become more complicated). Since there won't be thousands of instances of this, the little savings in the memory footprint would be irrelevant, so I'd avoid that.

...which raises the question: Can a device be both _faulty_ and _spare_ at the same time? Or both _faulty_ and _journal_? Probably not; but you never know how many more of those flags the future might bring, and multiplexing more and more of them into an enum might prove difficult when flags need to be added that are _not_ mutually exclusive.

shundhammer:
LGTM

PR #761, opened 2020-07-16
- probe dax flag for disks

aschnell:
Most of the PBI is about probing the DAX flag for disks - and most of that is updating the testsuite. The rest are minor code improvements.

Not for any PBI, feature or bug.


wfeldt:
lgtm

PR #762, opened 2020-07-20
- improve interface for pools

aschnell:
Improve the interface for pools by providing the functions Storage::get_pools() and Pool::size(devicegraph).

Not for any PBI, bug or feature.


storage/Storage.h
```diff
@@ -310,7 +310,15 @@ namespace storage
...
-	std::vector<std::string> get_devicegraph_names() const;
+	std::vector<std::string> get_devicegraph_names() const ST_DEPRECATED;
+
+	/**
+	 * Get all devicegraphs with their names.
+	 *
+	 * Note: There is no non-const version since that would provide non-const access
+	 * to the probed devicegraph.
+	 */
+	std::map<std::string, const Devicegraph*> get_devicegraphs() const;
```
joseivanlopez, line 321: NP: Why is this API (returning a map) needed? I mean, maybe it could simply return a vector of devicegraphs if a devicegraph provides a method to get its name.

aschnell: Well, the devicegraph does not know its name. And if the name was a member of the devicegraph the operator== would be strange as it would have to ignore the name (or be useless). Also having a map makes it clear that the name is unique.

PR #763, opened 2020-07-21
- probe lvm logical volume type from segtype for origins

aschnell:
For thick logical volumes the lvs command reports type 'o' (origin) if they have snapshots. In that case the type (normal, raid, cache, ...) must be probed by looking at the segtype field in the lvs output.

For https://trello.com/c/7modG3ZV/1916-2-improve-probing-of-lvm-snapshots.


wfeldt:
lgtm

PR #764, opened 2020-07-29
add _constraints file for OBS requiring at least 4 GB disk size (bsc#1174375)

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1174375

Building in OBS requires 4 GB disk size (at least on PPC).

Add a `_constraints` file ensuring this size (for all architectures).


aschnell:
LGTM

PR #789, opened 2021-01-07
Use non-deprecated location for efivars for EFI detection

dirkmueller:
Starting with Kernel 5.10 on aarch64, the deprecated firmware/vars
layout no longer exists. Since we only detect existence, we can
use firmware/efivars, which is the location that should work
on all architectures going forward.

### DIFF
--- a/storage/SystemInfo/Arch.cc
+++ b/storage/SystemInfo/Arch.cc
@@ -105,7 +105,7 @@ namespace storage
 	}
 	else
 	{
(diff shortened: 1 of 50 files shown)


aschnell:
Looks OK. In general references to documentation would be good.

ggardet:
> Looks OK. In general references to documentation would be good.

For the record, upstream kernel commit is: https://github.com/torvalds/linux/commit/963fabf37f6a94214a823df0a785e653cb8ad6ea

PR #796, opened 2021-01-22
Simplify compound actions for subvolumes

joseivanlopez:
### Problem

Compound actions for subvolumes are too complex because they inform about every change over the subvolume, for example:

Screenshot from 2021-01-21 15-47-03

Note that these sentences would become more and more complex when new attributes will be added to the subvolumes. 

### Solution

Simplify the compound actions for subvolumes in order to only show a short sentence when creating or editing a subvolume:

* *Create subvolume `name` on `device`*
* *Modify subvolume `name` on `device`*

Note that when a subvolume is edited, and only one action is done (e.g., activate 'nocow' property), its compound action would default to such single action text:

* *Set option 'no copy on write' for subvolume `name` on `device`*

Screenshot from 2021-01-28 13-42-32

### DIFF
--- a/storage/CompoundAction/Formatter/BtrfsSubvolume.h
+++ b/storage/CompoundAction/Formatter/BtrfsSubvolume.h
@@ -47,10 +47,9 @@ namespace storage
 
 	Text delete_text() const;
 
-	Text create_with_nocow_text() const;
-	Text create_with_limits_text() const;
-	Text create_with_nocow_and_limits_text() const;
 	Text create_text() const;
+
+	Text edit_text() const;
 
     private:
 
--- a/storage/CompoundAction/Formatter/BtrfsSubvolume.h
+++ b/storage/CompoundAction/Formatter/BtrfsSubvolume.h
@@ -49,7 +49,7 @@ namespace storage
 
 	Text create_text() const;
 
-	Text edit_text() const;
+	Text modify_text() const;
 
     private:
 
--- a/po/af.po
+++ b/po/af.po
@@ -5,7 +5,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: installation\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-01-20 09:42+0100\n"
+"POT-Creation-Date: 2021-01-28 14:51+0000\n"
 "PO-Revision-Date: 2018-02-02 15:00+0000\n"
 "Last-Translator: Arvin Schnell <aschnell@suse.com>\n"
 "Language-Team: Afrikaans <https://l10n.opensuse.org/projects/libstorage/ng-"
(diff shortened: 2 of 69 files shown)


aschnell:
Now the messages explain even less than they did before adding qgroup limits.

ancorgs:
Yes, they are less explanatory now but that aligns the subvolumes with the compound actions for other types of devices. In general we have compound actions like "Create partition sda1 for /home", instead of super-detailed actions like the ones in the actiongraph (which includes setting the partition id, setting the boot flag and whatsnot).

That was always the purpose of the compound actions, isn't it? Having sorter one-sentence summaries to display to the user in contrast with the detailed information available in the plain actions which detail every single property/attribute that is indeed modified.

aschnell:
Unfortunately we are going in circles. Originally there was a message for the btrfs to say that qgroups are modified: That was to simple. Then I made it more verbose: Not it now to complicated. This PR makes it partly less verbose than months ago. I consider that to be a very inefficient process.

aschnell:
Update to pot and po files is missing.

aschnell:
> [...] instead of super-detailed actions like the ones in the actiongraph (which includes setting the partition id, setting the boot flag and whatsnot).

If only the id is changed the messages does say that the id changes. For nocow of subvolumes that is now not the case anymore.

joseivanlopez:
> Unfortunately we are going in circles. Originally there was a message for the btrfs to say that qgroups are modified: That was to simple. Then I made it more verbose: Not it now to complicated. This PR makes it partly less verbose than months ago. I consider that to be a very inefficient process.

It was my fault because I introduced that "nocow" part in the sentence. Then, we wanted to have a compound action associated to the subvolume when its limit is set. But extending the sentence even more was an error again, and it was my fault once more (I suggested to do it without taking into account that it was not the original idea of compound actions).

After discussing it with @ancorgs, we came to the conclusion that the correct way is to have the very short sentences proposed here. After all, in the summary we only want to indicate that a subvolume is going to be created, edited or deleted, that's all.

> If only the id is changed the messages does say that the id changes. For nocow of subvolumes that is now not the case anymore.

Before the last PR, when a scenario is not controlled in the compound actions (like the fact of only changing the partition id), the first "single action" was used as default sentence. This had some problems, for example, when there was more than one single action. Now, all the "single actions" are used to compose the default value of a compound action (resulting in a verbose output). IMHO, we need to improve this. I would only use the "single action" as default (in uncontrolled scenario) when there is only one action associated to the compound action. Otherwise, I would use something like "Modify partition sda1".

ancorgs:
> > If only the id is changed the messages does say that the id changes. For nocow of subvolumes that is now not the case anymore.
> 
> Before the last PR, when a scenario was not controlled in the compound actions (like the fact of only changing the partition id), the first "single action" was used as default sentence. This had some problems, for example, when there were more than one single action. Now, all the "single actions" are used to compose the default value of a compound action (resulting in a verbose output). IMHO, we need to improve this. I would only use the "single action" as default (in uncontrolled scenario) when there is only one action associated to the compound action. Otherwise, I would use something like "Modify partition sda1".

That makes sense to me.

aschnell:
I think it is safer to have the default_text() function concat all actions. Otherwise the user might be fooled by only seeing one minor action, e.g. "set partition id", when the partition is also resized. And only presenting "modify partition" when it gets resized looks also bad to me.

joseivanlopez:
> I think it is safer to have the default_text() function concat all actions. Otherwise the user might be fooled by only seeing one minor action, e.g. "set partition id", when the partition is also resized. And only presenting "modify partition" when it gets resized looks also bad to me.

Maybe yes. Do you agree to a) keep default_text as it is now (with all actions) and b) simplify btrfs subvolume compound actions as proposed in this PR?

For a), I am not sure if we can simply add more compound actions "under demand". I mean, if we see that some common actions are resulting in a verbose output, then we can provide a new compound action. 

@aschnell, @ancorgs what do you think?

aschnell:
I can live with that though I do not see it as an improvement. Maybe some UX expert can help.

joseivanlopez:
> Update to pot and po files is missing.

@aschnell I have run `cd po; make merge` but git says that nothing has changed in pot and po files. How can I actually update them?

ancorgs:
Just evaluating the description and not really the code, it looks good to me.

The only concern I would have form the YaST point of view is that
"Create limits for qgroups of subvolume @/blah"
is more technical wording than the one we expose in the YaST UI. From that point of view and would prefer something closer to
"Modify quota of subvolume @blah".

But I can live with it, I guess.

storage/CompoundAction/Formatter/BtrfsSubvolume.h
```diff
@@ -47,11 +47,10 @@ namespace storage
 
 	Text delete_text() const;
 
-	Text create_with_nocow_text() const;
-	Text create_with_limits_text() const;
-	Text create_with_nocow_and_limits_text() const;
 	Text create_text() const;
 
+	Text edit_text() const;
```
aschnell, line 52: Please rename that to ```modify_text```. The actions are also called ```Create```, ```Modify``` and ```Delete```.

PR #838, opened 2021-10-14
Use dev name in fstab for LUKS devices (jsc#SLE-20416, bsc#1181196)

ancorgs:
## Problem

According to [bsc#1181196](https://bugzilla.suse.com/show_bug.cgi?id=1181196) (later moved to [jsc#PM-2460](https://jira.suse.com/browse/PM-2460) -> [jsc#SLE-20416](https://jira.suse.com/browse/SLE-20416)), specifying /etc/fstab entries via UUID= for LUKS devices shouldn't be done because it means that some tools, such as systemd generators, cannot know which LUKS device to activate to make a file-system appear unless all volumes are set up at boot.

Therefore instead of UUID, the reporter recommends to use the name of the resulting encrypted block device (eg. `/dev/mapper/cr_xxx `) since it also identifies the LUKS-backed device without ambiguity.

## Solution

This PR makes libstorage-ng use the device name by default for LUKS devices, instead of the global default `mount_by` method. That behavior is kind of equivalent to what libstorage-ng has always done for LVM LVs (which also have stable device names). 

This only changes the default `mount_by` proposed by libstorage-ng when creating a new mount point. It does not limit the options for the user to select any other mount_by method for the mount point (actually, using the default global value that is commonly UUID has worked for most cases in the past).

### DIFF
--- a/storage/Filesystems/BlkFilesystemImpl.cc
+++ b/storage/Filesystems/BlkFilesystemImpl.cc
@@ -37,6 +37,7 @@
 #include "storage/Holders/FilesystemUserImpl.h"
 #include "storage/Devices/BlkDeviceImpl.h"
 #include "storage/Devices/LvmLv.h"
+#include "storage/Devices/Luks.h"
 #include "storage/Devicegraph.h"
 #include "storage/SystemInfo/SystemInfo.h"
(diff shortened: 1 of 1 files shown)


aschnell:
Thanks for the patch.

PR #858, opened 2022-01-28
Add glibc-locale buildrequires for testsuite on SUSE distros

dirkmueller:
The testsuite uses non-base locales and hence needs the full glibc-locale package for working.


aschnell:
Looks fine.

Likely also needed for snapper.

PR #859, opened 2022-02-03
Modify mount point if mount_type changes

joseivanlopez:
## Problem

A mount point is not updated in */etc/fstab* if only its mount type changes.

## Solution

Consider the mount type when deciding if a mount point should be updated.

## Testing

Tested manually with YaST Expert Partitioner.

### DIFF
--- a/storage/Filesystems/MountPointImpl.cc
+++ b/storage/Filesystems/MountPointImpl.cc
@@ -442,7 +442,8 @@ namespace storage
 	if (lhs.in_etc_fstab && in_etc_fstab)
 	{
 	    if (lhs.mount_by != mount_by || lhs.mount_options != mount_options ||
-		lhs.freq != freq || lhs.passno != passno || lhs.path != path)
+		lhs.freq != freq || lhs.passno != passno || lhs.path != path ||
+		lhs.mount_type != mount_type)
 	    {
 		actions.push_back(make_shared<Action::UpdateInEtcFstab>(get_sid()));
 	    }
--- a/storage/Filesystems/MountPointImpl.cc
+++ b/storage/Filesystems/MountPointImpl.cc
@@ -432,9 +432,9 @@ namespace storage
 	}
 
 	// A unmount action could be required when the device is set as unmounted in the target system
-	// or when its mount path has changed. But the unmount action only makes sense if the device is
-	// currently mounted in the system.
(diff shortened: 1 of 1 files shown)


aschnell:
There is an integration test for that (https://github.com/openSUSE/libstorage-ng/blob/master/integration-tests/mount-points/set-mount-by.py) and it works for me without the patch.

Trying to change it in YaST I get a error popup about the mount point being used already.

joseivanlopez:
> There is an integration test for that (https://github.com/openSUSE/libstorage-ng/blob/master/integration-tests/mount-points/set-mount-by.py) and it works for me without the patch.

That test is about the mount_by, right? This change is to consider the mount_type when editing the mount point.
 
> Trying to change it in YaST I get a error popup about the mount point being used already.

Where do you get the error, during commit phase? BTW, to check this fix, you have to edit a NFS with old nfs4 in the type field of the fstab. YaST will replace nfs4 by nfs.

aschnell:
Sorry, I mixed mount-by and mount-type.

But isn't strictly speaking remounting required to go from nfs to nfs4? And AFAIS this needs a unmount and mount.

joseivanlopez:
> Sorry, I mixed mount-by and mount-type.
> 
> But isn't strictly speaking remounting required to go from nfs to nfs4? And AFAIS this needs a unmount and mount.

Yes, but YaST keeps the #active status of the mount point, but it automatically fixes the type field in the fstab if needed. So, it is possible that YaST modifies the fstab entry without unmounting/mounting the device.

aschnell:
What I mean: The library should do the unmount/mount like it does already when the path is changed. It is not included in the patch but should be easy. In ```MountPoint::Impl::add_modify_actions``` just twice a few lines below the current addition.

joseivanlopez:
Ah, now I got it! But I have a question. Should the library unmount/mount only if `path` or `mount_type` changes? Should the `mount_options` be considered too?

aschnell:
When changing ```mount-by``` a remount does not look needed to me. Why which udev link the devices is mounted should be irrelevant once it is mounted. It isn't even visible in /proc/mounts.

Changing the ```mount options``` is more interesting. Yes, a remount should be done. But depending on the options a real remount instead of unmount and mount should be done (mount(2) lists the options supported by remount). So I would handle that as a different issue.

aschnell:
Looks fine. Thanks,

PR #870, opened 2022-04-26
LUKS2: add AEAD integrity option

aplanas:
LUKS2 support "integrity" and "data authentication" since 2.0 in
experimental mode.  This is required when we want to detect changes out
of band in the devices.

Recently the kernel is supporting good AEAD family of algorithms that
will guarantee unauthorized changes in the device (instead of decrypting
random data) for big disks.

For integrity it is recommended "aegis128-random" with AEAD, as other
family (like chacha20-random with poly1305, or aes-xts-plain64 with
hmac-sha256) have performance problems, or are not supporting large
enought devices.  A check is implemented to validate the data.

This patch add a new element, "integrity", that for now can only have
"aead" value.  When other methods are added, the check can be changed
without changing the XML data, nor the parser.  If integrity is set,
also it is checked that "aegis128-random" cipher algorithm is used.

Signed-off-by: Alberto Planas <aplanas@suse.com>


aschnell:
Is the resulting device with integrity smaller than without? In that case the space calculation must be extended.

storage/Devices/Encryption.h
```diff
@@ -143,6 +143,24 @@ namespace storage
 	 */
 	void set_pbkdf(const std::string& pbkdf);
 
+	/**
+	 * Get the integrity. For now we only expect AEAD.
+	 *
+	 * Currently only supported for LUKS2.
```
aschnell, line 149: Please add that it is even experimental in cryptsetup/luks.

aplanas: done

aplanas:
> Is the resulting device with integrity smaller than without? In that case the space calculation must be extended.

Do I need to implement some method? Can you point me to some example in the code?

Would be OK if the estimation is an approximation?

aplanas:
> Luks::Impl::calculate_region_and_topology()

Thanks. Should be done.

aplanas:
> Please add a testcase for probing in testsuite/probe.

I will try to do. So far there are no version 2 luks tests there, so I need to create all the fixtures from scratch.

storage/Devices/LuksImpl.cc
```diff
@@ -603,6 +617,14 @@ namespace storage
 		ST_THROW(Exception("invalid encryption type"));
 	}
 
+	if (get_cipher() == "aegis128-random" || !get_integrity().empty()) {
+	    if (get_cipher() != "aegis128-random")
+		ST_THROW(Exception("when integrity is set it is expected aegis128-random cipher"));
+
+	    if (get_integrity() != "aead")
+		ST_THROW(Exception("for now only AEAD integrity is supported"));
+	}
```
aschnell, line 626: Checking this during commit is very late. Likely a better place is in check().

And please add a document explaining the situation (what works, what does not work, what kernel is needed, ...) in the doc directory.

storage/SystemInfo/CmdCryptsetup.h
```diff
@@ -103,6 +104,11 @@ namespace storage
 	 */
 	string pbkdf;
 
+	/**
+	 * Is integrity used. Currently is only expected AEAD. Only fir LUKS2.
```
aschnell, line 108: Typo.

aplanas: Fixed.

aplanas:
> If probing works you can just save the mockup and devicegraph using utils/probe --save-mockup --save-devicegraph.

Done. There is a new test in prob using this process (integrity-*)

I think that I addressed all the comments.

aschnell:
cryptsetup works for me only when also setting --key-size 128. If that is expected please document it. And create a integration test which does test the creation.

aplanas:
> cryptsetup works for me only when also setting --key-size 128. If that is expected please document it.

Right. I will add it.

> And create a integration test which does test the creation.

I will check how to do that.

aschnell:
There is also a problem with probing due to the extra dm table. If the LUKS device is open the probed name includes the "_dif". Here the code Luks::Impl::probe_lukses() must check the target and if it is "integrity" find the table with target "crypt".

The probe test right now does not check this since the LUKS device is not open.

aschnell:
Is QA going to test this? And reference to a feature request?

aplanas:
> Is QA going to test this?

I cannot answer this, but I guess so.

> And reference to a feature request?

It is part of FDE and PM-3419

aplanas:
> There is also a problem with probing due to the extra dm table. If the LUKS device is open the probed name includes the "_dif". Here the code Luks::Impl::probe_lukses() must check the target and if it is "integrity" find the table with target "crypt".

I will check what it this is about. But did not the added test with the mocks was checking this?

> The probe test right now does not check this since the LUKS device is not open.

integration-tests/luks/create-with-integrity.py
```diff
@@ -0,0 +1,35 @@
...
+environment = Environment(False)
+
+storage = Storage(environment)
+storage.probe()
+
+staging = storage.get_staging()
+
+sdc1 = Partition.find_by_name(staging, "/dev/sdc1")
+
+luks = to_luks(sdc1.create_encryption("cr-test"))
```
aschnell, line 21: The create_encryption function without providing the encryption type is deprecated.

aplanas: Done.

storage/Devices/LuksImpl.cc
```diff
@@ -400,7 +419,31 @@ namespace storage
...
+			if (table.target == "integrity")
+			    integrity = true;
+		    }
+
+		    if (integrity)
+		    {
+			dm_table_name = dm_table_name.substr(0, dm_table_name.length() - 4);
+			const CmdDmsetupInfo& cmd_dmsetup_info = system_info.getCmdDmsetupInfo();
+			if (!cmd_dmsetup_info.exists(dm_table_name))
+			    ST_THROW(Exception("integrity device found, missing the associated crytpo target"));
```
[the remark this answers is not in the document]
aplanas: Not sure to understand. "crypt" is the name in the mockup, but can be anything. If there is an integrity device with _dif, there must be an associated device with target "crypt", that is why was like that before.

storage/Devices/LuksImpl.cc
```diff
@@ -400,7 +419,31 @@ namespace storage
...
+			if (table.target == "integrity")
+			    integrity = true;
+		    }
+
+		    if (integrity)
+		    {
+			dm_table_name = dm_table_name.substr(0, dm_table_name.length() - 4);
+			const CmdDmsetupInfo& cmd_dmsetup_info = system_info.getCmdDmsetupInfo();
+			if (!cmd_dmsetup_info.exists(dm_table_name))
+			    ST_THROW(Exception("integrity device found, missing the associated 'crytp' target"));
```
aschnell, line 443: I am just talking about the typo here: "crytp" -> "crypt"

aplanas: I am blind. Done. (Sorry)

aplanas:
> Please update the probe test to have the LUKS device open so that the new code in Luks::Impl::probe_lukses is used.

Right. Done.

storage/Devices/LuksImpl.cc
```diff
@@ -96,6 +107,14 @@ namespace storage
 	calculate_region_and_topology();
     }
 
+    void
+    Luks::Impl::set_integrity(const string& integrity)
+    {
+	Encryption::Impl::set_integrity(integrity);
+
+	if (get_integrity() != integrity)
```
aschnell, line 115: Is that ever the case?

aplanas: In reality no, but maybe someone put manually this suffix?

storage/Devices/LuksImpl.cc
```diff
@@ -496,6 +541,10 @@ namespace storage
 	else
 	    size = 0 * B;
 
+	// If we use integrity, calculate a pesimistic size
+	if (!get_integrity().empty())
+	    size = static_cast<unsigned long long>(size * 0.85);
```
aschnell, line 546: Why not ```size *= 0.85```?

aplanas: To be more explicit in the type, but I will change it (not used to C++ idioms)

storage/Devices/LuksImpl.cc
aschnell: After ```set_integrity(integrity)``` the condition is always false AFAIS.

aplanas: Ah right, this will never call the reize. I will move the condition before.

aplanas: Done. If you think that the condition should be dropped, we need to figure out why the tests are falling.

storage/Devices/LuksImpl.cc
aplanas: Done

storage/Devices/Encryption.h
```diff
@@ -143,6 +143,24 @@ namespace storage
...
+	 * Get the integrity. For now we only expect AEAD.
+	 *
+	 * Currently only supported for LUKS2 (experimental).
+	 */
+	const std::string& get_integrity() const;
+
+	/**
+	 * Set the integrity. If the integrity is set, this will
+	 * restrict the number of allowed cipher algorithms.  Also the
+	 * expected value in "aead".  Technically this can be replaced
```
aschnell, line 156: "in" -> "is". Empty is also allowed.

storage/Devices/Encryption.h
```diff
@@ -143,6 +143,24 @@ namespace storage
...
+	 * as a bootlean flag, but we maintain a string in case of
```
aschnell, line 157: "bootlean" -> "boolean".

storage/Devices/Encryption.h
```diff
@@ -143,6 +143,24 @@ namespace storage
...
 
+	/**
+	 * Get the integrity. For now we only expect AEAD.
+	 *
+	 * Currently only supported for LUKS2 (experimental).
+	 */
+	const std::string& get_integrity() const;
+
+	/**
+	 * Set the integrity. If the integrity is set, this will
```
aschnell, line 154: Like other functions (set_pbkdf) it should mention that the value is "Only used during creation.".

aplanas: Done

storage/Devices/Encryption.h
aplanas: Fixed.

storage/Devices/Encryption.h
aplanas: Fixed.

storage/Devices/LuksImpl.cc
aschnell: Likely a check like in ```Md::Impl::calculate_region_and_topology()``` is needed:

```
	if (exists_in_system())
	    return;
```

Explanation is in Md::Impl.

aplanas: Added check in local `calculate_region_and_topology()`

aschnell:
Looks good now. Please add the feature request in the git message and increase version in LIBVERSION to .1.74.0.

aplanas:
> Looks good now. Please add the feature request in the git message and increase version in LIBVERSION to 1.74.0.

Should be done.

PR #884, opened 2022-06-22
- handle rootprefix when reading fstab, crypttab and mdadm.conf

aschnell:
Needed for https://github.com/yast/yast-in-container. Further changes are needed.


aschnell:
I also did some tests and LVM seems to have a problem with the setup: The device nodes for logical volumes does not appear (could be a problem of udev).

Another problem I noticed is that mounting a file system in the container works but of course it is not mounted in the host as the user would expect.

PR #893, opened 2022-08-01
- support mounting by partition uuid and label

aschnell:


ancorgs:
What's the scope of this? I mean, is there any feature request to add partuuid and partlabel to SLE?

Asking to decide how to adapt YaST.

aschnell:
The team behind TIU (https://github.com/thkukuk/tiu) asked about mounting using PARTLABEL. There are no other request that I know of. So I assume YaST does not have to provide options to use those new mount-by types.

PR #900, opened 2022-10-18
- make result of ParitionTable::is_partition_id_supported() depend on parted version

aschnell:


ancorgs:
This PR broke this defensive test at yast-storage-ng: https://github.com/yast/yast-storage-ng/blob/master/test/y2storage/partition_tables/msdos_test.rb#L54

Is really expected that UNKNOWN is accepted in an MS-DOS partition table.

aschnell:
ID_UNKNOWN should only be set as a result of probing, not by e.g. set_id(). That is the same as with GPT and DASD partition tables.

Since on MS-DOS partition tables the id is only a single byte and all values can be represented, the id should never be UNKNOWN.

ancorgs:
@aschnell Since for YaST we keep SUSE:SLE-15-SP5:GA and TW in sync. It would be nice if you would submit this to SLE-15-SP5 as well.

aschnell:
Sure, SR created.

PR #905, opened 2022-11-25
- limit allowed by-id links for NVMEs (bsc#1205352)

aschnell:
See https://bugzilla.suse.com/show_bug.cgi?id=1205352.


data/udev-filters.json
```diff
@@ -0,0 +1,21 @@
+{
+    "Disk" : {
+	"allowed-path-patterns" : [ "ccw-*", "scsi-*", "pci-*", "platform-*" ],
+	"allowed-id-patterns" : [ "ata-*", "scsi-*", "wwn-*",
```
mwilck, line 4: `scsi-*` links are not necessarily unique. The best ones are `scsi-3*`, `scsi-2*`, `scsi-8*`. Less good: `scsi-1*`. Not recommended: `scsi-0*`, `scsi-S*`.

`wwn-*` is better, it's basically the same as `scsi-3*`.

But if you want to support legacy hardware, you should not exclude the non-unique `scsi-*` links entirely, but fall back to them if no better ones are available.

data/udev-filters.json
```diff
@@ -0,0 +1,21 @@
...
+	"allowed-id-patterns" : [ "ccw-*" ]
+    },
+    "Md" : {
+	"allowed-id-patterns" : [ "md-uuid-*" ]
+    },
+    "DmRaid" : {
+	"allowed-id-patterns" : [ "raid-*" ]
+    },
+    "Multipath" : {
+	"allowed-id-patterns" : [ "scsi-*", "wwn-*", "dm-name-*", "dm-uuid-*" ]
```
mwilck, line 19: What exactly do you need these for? For multipath devices, we recommend 
`dm-uuid-mpath-*` and for partitions on multipath devices `dm-uuid-part*-mpath-*`. See also the [SLE storage guide, §18.2.4](https://documentation.suse.com/sles/15-SP4/html/SLES-all/cha-multipath.html#sec-multipath-referring).

If you need to use the lower-layer links (like `scsi-*`) here, you may also need to include the `ccw-*` and possible `nvme-*` links.

mwilck (commented):
So you went for a "whitelist" kind of approach. Fine with me, but I thought you were aiming for some sort of priority algorithm for links with different "reliability" or "uniqueness".

data/udev-filters.json
mwilck: Just saw that you were asking for a separate bug report for this. I'll see what I can do, no promises here.

PR #923, opened 2023-04-05
detect transport for NVMe disks (bsc#1210144)

aschnell:
See https://bugzilla.suse.com/show_bug.cgi?id=1210144.


storage/Devices/DiskImpl.cc
```diff
@@ -152,6 +153,20 @@ namespace storage
     }
 
 
+    string
+    Disk::Impl::get_nvme_controller() const
+    {
+	static const regex nvme_name_rx(DEV_DIR "/nvme([0-9]+)n([0-9]+)", regex::extended);
+
```
mwilck, line 160: I don't think this will always work. In particular, with NVMe native multipathing, the first integer matches the **subsystem ID**, not the controller ID. 

The correct way to determine the controllers that provide connections for a given namespace (disk) works like this:
```
# readlink -f /sys/class/block/nvme4n1/device
/sys/devices/virtual/nvme-subsystem/nvme-subsys4
# find /sys/devices/virtual/nvme-subsystem/nvme-subsys4 -regex '.*/nvme[0-9]*' -printf '%f -> %l\n'
nvme7 -> ../../nvme-fabrics/ctl/nvme7
nvme5 -> ../../nvme-fabrics/ctl/nvme5
nvme6 -> ../../nvme-fabrics/ctl/nvme6
nvme4 -> ../../nvme-fabrics/ctl/nvme4
```

Note that in this real world example, the controller `nvme4` actually belongs to `nvme-subsys4`, but that's not necessarily the case. The numbers can, and will, be different.

It's even possible that a namespace could be connected with several controllers using different transports.

Sorry, I should have noted this on bugzilla more clearly.

mwilck (commented):
See remark about controller numbering vs subsystem/namespace numbering.

storage/Devices/DiskImpl.cc
aschnell: How realistic is the case of several different transports for one device? It is just theoretic or something seen in the field?

mwilck: It is unrealistic, but not impossible. I really can't tell. For practical purposes, it's probably sufficient to just return the transport of the first path in the multipath case. After all, the most important question is whether it's PCIe or (some sort of) fabrics, and I'm pretty sure that those two can't be mixed. (I don't know for certain if NVMe / PCIe supports multipath at all).

aschnell: I think it is better to use output of ```nvme list --verbose --output json``` than looking for so many files in /sys since it should be easier to extend and much better to debug.

I have the output of two systems. The simple one (Notebook with NVMe) does not list any namespaces. So I would have to assume that "/dev/nvme0n1" corresponds to the subsystem "nvme-subsys0". Is that always the case?

mwilck: The numbering depends on the setting of the `nvme_core.multipath` parameter. If it is true (default both upstream and on SLE15 and later), the first integer is the subsystem ID. Otherwise, it is the controller ID.

See [the kernel source code](https://elixir.bootlin.com/linux/v6.2.11/source/drivers/nvme/host/core.c#L4290). The comment explain the rationale for this complex numbering scheme.

The second integer is always namespace ID (also a kernel ida number, not necessarily the same as the hardware NSID — but I don't think you need to bother about that).

PR #936, opened 2023-06-15
- use correct error callback when nvme-cli is missing during probing

aschnell:
For https://github.com/yast/yast-storage-ng/issues/1343.


shundhammer (approved):
LGTM, but I'd be more comfortable if Arvin had a look at it as well.

PR #950, opened 2023-10-09
- allow to disable use of key file in commit

aschnell:


storage/Devices/EncryptionImpl.cc
```diff
@@ -424,7 +431,7 @@ namespace storage
 
 	SystemCmd::Options cmd_options(cmd_line, SystemCmd::DoThrow);
 
-	if (!get_key_file().empty())
+	if (!get_key_file().empty() && use_key_file_in_commit)
```
ancorgs, line 434: Just a dumb question out of curiosity. Why not using `is_use_key_file_in_commit()` here? I just found a bit strange using the getter for `key_file` in the first half of the condition and using the attribute `use_key_file_in_commit` directly in the second half.

aschnell: There is no reason for that, except my inability to make it always consistent.

PR #967, opened 2023-12-07
block more udev by-id links (bsc#1217459)

aschnell:


mwilck:
The general solution should work.

I am not sure why you removed the unreliable identifiers in all test cases. IMO libstorage-ng should identify these devices correctly, whether or not the additional identifiers are present. Note that customers/users can configure the creation of these symlinks in order to provide a simple workaround for cases in which they might be necessary (https://en.opensuse.org/SDB:SCSI_Device_Identification).

aschnell:
I had to remove the unreliable identifiers in the test cases: These tests read the mockup files (which e.g. include the output of udevadm info including the unreliable identifiers), simulate the probing and checks whether the resulting devicegraph matches the expected. Since the probing drops the links I had to modify the expectation.

PR #971, opened 2023-12-21
- adapted to changed nvme json output (bsc#1218306)

aschnell:


DimStar77:
Technically, libstorage-ng should now conflict with nvme-cli < 2.7 ( to ensure we don't run this against nvme-cli 2.6 where HostID is used)

DimStar77:
And take into account https://github.com/linux-nvme/nvme-cli/issues/2166 - this is supposedly being reverted in nvme

igaw:
Yes, the idea is that we have a stable JSON output. Thus I rather revert all field name changes in nvme-cli and you don't have to touch your parser.

igaw:
I verify that `nvme list-subsys -o json` will give out the same field names as with v2.6. I'll also double check if any other field name has changed. Sorry for the mess.

aschnell:
OK, surely I changed the parser to handle the old and new text. But fixing it in nvme is even better.

PR #1090, opened 2026-08-26
Fix SWIG 4.5.0 template instantiations and logger crashes

DimStar77:
Surgically resolve all integration and runtime build errors arising from
    the SWIG 4.5.0 upgrade while maintaining backwards compatibility with older
    SWIG versions:
    
    1. Disable -autorename command-line flag in bindings/ruby/Makefile.am:
       Under SWIG 4.5.0, -autorename causes a redefinition conflict that silently
       discards all pointer vector templates (like VectorLvmVgPtr), preventing
       proper Ruby array conversions. Disabling it also resolves C++ const-correctness
       overload resolution failures (such as with to_lvm_vg).
    
    2. Add RubyLogger GC guard and specific logger= rename:
       Wrap set_logger in a custom RubyLogger to prevent C++ thread logging from
       allocating Ruby objects during GC phases, which crashes the VM. Place the
       %rename("logger=") before regex patterns in storage-ruby.i to avoid shadowing,
       and define set_ruby_logger cleanly in storage.i.

https://bugzilla.opensuse.org/show_bug.cgi?id=1275515

### DIFF
--- a/bindings/ruby/Makefile.am
+++ b/bindings/ruby/Makefile.am
@@ -30,7 +30,7 @@ storage_la_DEPENDENCIES =				\
 	${top_builddir}/storage/libstorage-ng.la
 
 storage_wrap.cxx: storage-ruby.i $(DEPENDENCIES)
-	swig -o storage_wrap.cxx -c++ -ruby -autorename ${AM_CPPFLAGS} storage-ruby.i
+	swig -o storage_wrap.cxx -c++ -ruby ${AM_CPPFLAGS} storage-ruby.i
 
 CLEANFILES = storage.so storage_wrap.cxx storage_wrap.h
 
--- a/bindings/ruby/storage-ruby.i
+++ b/bindings/ruby/storage-ruby.i
@@ -8,6 +8,8 @@
 
 %rename("empty?") "empty";
 
+%rename("logger=") storage::set_ruby_logger;
+
 %rename("%(regex:/^(get_)(.*)/\\2/)s") "";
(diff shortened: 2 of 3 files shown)


aschnell:
Thanks for looking into this.

Removing ```-autorename``` is fine for me. Surely it does change some function and class names for Ruby. E.g. ```Exception::as_string()``` will become ```Exception::asString()```. Here an "alias" could be added in C++. Some names were just renamed strange, e.g. ```to_f2fs()``` was renames to ```to_f_2fs()``` (why on earth?). I checked some of those functions and could not find uses in yast2-storage-ng. But there could be regressions.

Concerning the logging: When does this happen? During termination of YaST? Did it happen in a testsuite? What I do not like about the patch is that in a multi-threaded context ```rb_during_gc()``` could return false and GC could start before ```write``` is called. But I do not know enough about Ruby to say whether that can happen.

Apart from that the documentation of ```set_logger()``` states that the logger object has to be valid until a new one is set. But during abnormal program termination that can be difficult to ensure.

DimStar77:
the GC crash was observed in the test suite (don't recall exactly if of this package of yast2-storage-ng, which was the main consumer tested against this change)

DimStar77:
oh - and as noted on the bug report: this PR was 120% Gemini; I do not claim to understand swig :)

aschnell:
OK, I have reproduced the crash in OBS. I will merge the PR now and then make a few changes. E.g. the ruby swig code could be moved to bindings/ruby/storage-ruby.i.

